PR #9, opened 2025-11-22
Add MCP Ping Support with Test Infrastructure Improvements

abigagli:
Implements the MCP `ping` utility method as specified in the [Model Context Protocol specification](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/ping), along with fixes to the test infrastructure that improve reliability across the test suite.

#### Ping Handler
- Added `ping` method handler in `src/mcp/server.rs`
- Returns empty object `{}` per MCP specification requirements
- Stateless, synchronous operation with no parameters
- Enables clients to verify server responsiveness and measure round-trip latency

#### Test Infrastructure Fixes

**Socket Blocking Mode Fix:**
Fixed a critical issue in the test IPC server where accepted connections inherited non-blocking mode from the listener socket, causing spurious `EAGAIN`/`EWOULDBLOCK` errors during request processing.

*Root Cause:*
- The listener socket uses non-blocking mode to enable graceful shutdown (checking `should_shutdown` between `accept()` calls without blocking indefinitely)
- On Unix systems, `accept()` returns connected sockets that inherit the listener's non-blocking flag (POSIX behavior)
- When the server attempts to read requests via `BufReader::read_line()`, non-blocking mode causes reads to fail with `WouldBlock` if data isn't immediately available in the kernel buffer
- This manifested as intermittent failures on sequential requests, where timing variations in pipe/socket buffering would cause reads to fail even though data was in transit

*Solution:*
Explicitly set accepted connections to blocking mode with `stream.set_nonblocking(false)` immediately after `accept()`. This allows the listener to remain non-blocking for shutdown responsiveness while ensuring connected sockets reliably wait for complete request data.

**Additional Infrastructure Improvements:**
- **Path Canonicalization**: Added workspace path canonicalization in `test-support/src/ipc/client.rs` to ensure consistent absolute path handling
- **Unix Socket Path Length**: Shortened socket paths to comply with SUN_LEN limits (104 bytes on macOS) by using abbreviated names
- **Timeout Support**: Added configurable timeout to `IpcClient::send_request` for better failure detection

#### Test Coverage

**Integration Tests** (`tests/integration/mcp_server_test.rs`):
- Response format validation
- Idempotency verification (5 sequential pings)
- Rapid-fire throughput testing (10 sequential pings)
- Post-ping server responsiveness verification

**Stress Tests** (`tests/stress/concurrent_requests.rs`):
- `test_ping_rapid_fire`: Sequential throughput (100-500 requests)
- `test_ping_concurrent`: Concurrent request handling (10-50 parallel)
- `test_ping_interleaved_with_tools`: Ping/tool call alternation (20-50 iterations)
- `test_ping_mixed_concurrent_workload`: Mixed concurrent workload with 40% ping ratio (15-30 concurrent)
- `test_ping_burst_pattern`: Sustained burst patterns (5-10 bursts of 20-50 requests)
- `test_ping_connection_stability`: Connection lifecycle testing (50-100 iterations)

All tests include CI-specific tuning with reduced iteration counts for GitHub Actions environment.

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -1,5 +1,5 @@
 use anyhow::{anyhow, Result};
-use log::info;
+use log::{info, warn};
 use serde_json::{json, Value};
 use std::{
     collections::{HashMap, HashSet},
@@ -356,6 +356,20 @@ impl RustAnalyzerClient {
         self.diagnostics.lock().await.clear();
         self.initialized = false;
         Ok(())
+    }
+}
+
+impl Drop for RustAnalyzerClient {
+    fn drop(&mut self) {
+        if let Some(ref mut process) = self.process {
+            info!("Killing child rust-analyzer process on drop");
+            if let Err(e) = process.start_kill() {
+                warn!("Failed to kill rust-analyzer process on drop: {}", e);
+            }
+            // Best-effort reap to avoid zombies. If the process hasn't exited yet,
+            // the brief zombie will be reaped by init/launchd when we exit shortly after.
+            let _ = process.try_wait();
+        }
     }
 }
 
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -86,15 +86,35 @@ impl RustAnalyzerMCPServer {
         let mut reader = BufReader::new(stdin);
         let mut writer = BufWriter::new(stdout);
 
-        // Handle shutdown signals.
+        // Handle shutdown signals. SIGINT and SIGTERM are handled in separate
+        // tasks intentionally. Using tokio::select! to combine them in a single
+        // task causes the losing future to be dropped/cancelled, which can leave
+        // tokio's internal signal handler state inconsistent. This manifests as
+        // the MCP process exiting unexpectedly when run as a child process (e.g.
+        // by the test-support-server), causing broken pipe errors in tests.
+        // Two independent tasks avoid any cross-signal interference.
         let running = Arc::new(Mutex::new(true));
         let running_clone = Arc::clone(&running);
 
         tokio::spawn(async move {
             let _ = tokio::signal::ctrl_c().await;
-            info!("Received shutdown signal");
+            info!("Received SIGINT");
             *running_clone.lock().await = false;
         });
+
+        // Also handle SIGTERM (sent by process managers and editors like Windsurf).
+        #[cfg(unix)]
+        {
+            let running_clone2 = Arc::clone(&running);
+            tokio::spawn(async move {
+                use tokio::signal::unix::{signal, SignalKind};
+                let mut sigterm =
+                    signal(SignalKind::terminate()).expect("Failed to register SIGTERM handler");
+                sigterm.recv().await;
+                info!("Received SIGTERM");
+                *running_clone2.lock().await = false;
+            });
+        }
 
         loop {
             // Check if we should stop.
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -158,6 +158,11 @@ impl RustAnalyzerMCPServer {
                     }
                 }),
             },
+            "ping" => MCPResponse::Success {
+                jsonrpc: "2.0".to_string(),
+                id: request.id,
+                result: json!({}),
+            },
             "tools/list" => MCPResponse::Success {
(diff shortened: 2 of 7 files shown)


test-support/src/ipc/client.rs
```diff
@@ -38,9 +38,58 @@ impl IpcClient {
...
+                            request_id: AtomicU64::new(1),
+                            workspace_path,
+                        });
+                    }
+                }
+            }
+        }
+
+        // Re-check after acquiring the lock: another process may have started
+        // the server while we were waiting.
```
zeenix, line 92: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] The success path immediately below returns while leaving the newly-created `.lock` file behind. Once the server later exits, every future caller sees `AlreadyExists` and spins forever because no server can become reachable; build/spawn error paths leak the lock too. Please use cleanup that runs on every exit (and handles stale owners), or preferably drop this commit because current `main` already arbitrates startup by binding the socket before rust-analyzer initialization.

test-support/src/ipc/client.rs
```diff
@@ -38,9 +38,58 @@ impl IpcClient {
...
+        // test-and-set: only one process succeeds; the others spin-wait.
+        let lock_path = socket_path(project_type).with_extension("lock");
+        loop {
+            match std::fs::OpenOptions::new()
+                .write(true)
+                .create_new(true)
+                .open(&lock_path)
+            {
+                Ok(_guard) => break, // We hold the lock.
+                Err(_) => {
```
zeenix, line 73: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Only `ErrorKind::AlreadyExists` means another starter owns this path. All other errors need to be returned; treating them as contention makes permanent filesystem failures spin forever.

src/mcp/server.rs
```diff
@@ -86,16 +86,36 @@ impl RustAnalyzerMCPServer {
...
+        #[cfg(unix)]
+        {
+            let running_clone2 = Arc::clone(&running);
+            tokio::spawn(async move {
+                use tokio::signal::unix::{signal, SignalKind};
+                let mut sigterm =
+                    signal(SignalKind::terminate()).expect("Failed to register SIGTERM handler");
+                sigterm.recv().await;
+                info!("Received SIGTERM");
+                *running_clone2.lock().await = false;
```
zeenix, line 115: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Flipping this flag does not wake the loop’s pending `read_line`, so SIGINT/SIGTERM can still leave the server hung until stdin produces another byte or EOF. Current `main` now races reads, request handling, and writes against a persistent `ShutdownSignal`; please drop this superseded signal implementation when rebasing.

src/lsp/client.rs
```diff
@@ -359,6 +359,20 @@ impl RustAnalyzerClient {
...
+impl Drop for RustAnalyzerClient {
+    fn drop(&mut self) {
+        if let Some(ref mut process) = self.process {
+            info!("Killing child rust-analyzer process on drop");
+            if let Err(e) = process.start_kill() {
+                warn!("Failed to kill rust-analyzer process on drop: {}", e);
+            }
+            // Best-effort reap to avoid zombies. If the process hasn't exited yet,
+            // the brief zombie will be reaped by init/launchd when we exit shortly after.
+            let _ = process.try_wait();
```
zeenix, line 371: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: `start_kill()` only initiates termination; an immediate `try_wait()` may return `Ok(None)`, after which dropping the handle does not guarantee the child was reaped. Please arrange async kill+wait on all normal/error paths and use `kill_on_drop(true)` only as the final backstop.

zeenix (commented):

The ping implementation itself looks fine, but this is no longer an atomic ping PR: the first commit mixes ping and IPC behavior, and the branch now also contains #14 plus independent test-startup and process/signal fixes, some superseded by current `main`. Please rebase and rewrite #9 to the ping handler plus a focused test, and move any surviving test-support changes into separately prefixed atomic commits/PRs as `CONTRIBUTING.md` requests. All six current subjects also lack the requested package prefix; the final three lack emoji prefixes.

PR #10, opened 2025-12-03
Fix request handler for notifications

ju6ge:
json rpc notifications are sent without a request id and must not be
responded to. This fix updates the logic to simply accept notifications
without generating a response. Sending a response can lead to crashes
within mcp clients and render the mcp server unusable

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -126,11 +126,15 @@ impl RustAnalyzerMCPServer {
             };
 
             debug!("Received request: {}", request.method);
-            let response = self.handle_request(request).await;
-            let response_json = serde_json::to_string(&response)?;
-            writer.write_all(response_json.as_bytes()).await?;
-            writer.write_all(b"\n").await?;
-            writer.flush().await?;
+
+            // requests without an id are notifications and must not receive a response!
+            if request.id.is_some() {
+                let response = self.handle_request(request).await;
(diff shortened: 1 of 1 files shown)


src/mcp/server.rs
```diff
@@ -126,11 +126,16 @@ impl RustAnalyzerMCPServer {
             };
 
             debug!("Received request: {}", request.method);
-            let response = self.handle_request(request).await;
-            let response_json = serde_json::to_string(&response)?;
-            writer.write_all(response_json.as_bytes()).await?;
-            writer.write_all(b"\n").await?;
-            writer.flush().await?;
+            log::debug!("{request:#?}");
```
zeenix, line 129: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This full-payload log is unrelated to the notification fix, duplicates the method log above, and can expose all file paths and tool arguments in debug output. Please drop commit `5d1d692`; if full request logging is wanted, it should be proposed separately with its own rationale.

zeenix (commented):

The notification bug is real, but #15 is the smaller canonical fix and #17 already contains its exact commit. I recommend closing this as superseded after #15 lands. If this PR is continued instead, it must be rebased onto current `main` and the unrelated logging below should be dropped. The commit subjects also need emoji/package prefixes, and the second commit needs a detailed body, per `CONTRIBUTING.md`.

PR #11, opened 2026-01-06
feat: Add new LSP tools and robustness improvements

panosAthDBX:
## Summary

This PR adds several new MCP tools and improves the robustness of the rust-analyzer integration.

### New Tools

| Tool | Purpose | LSP Method |
|------|---------|------------|
| `workspace_symbols` | Search symbols across entire workspace | `workspace/symbol` |
| `implementations` | Find trait implementations | `textDocument/implementation` |
| `incoming_calls` | What calls this function? | `callHierarchy/incomingCalls` |
| `outgoing_calls` | What does this function call? | `callHierarchy/outgoingCalls` |

### Enhanced Tools

- **`symbols`**: Added `include_hover` parameter to fetch type signatures and documentation for each symbol in one call

### Robustness Improvements

- **Crash Detection & Auto-Restart**: Automatically restarts rust-analyzer up to 3 times if it crashes
- **LRU Document Management**: Evicts oldest documents when capacity (50) is reached to prevent memory growth
- **Request Cancellation**: Sends `$/cancelRequest` to rust-analyzer on timeout
- **Notification-Based Diagnostics**: Uses `tokio::sync::Notify` for efficient diagnostic updates

### Dependency Updates

| Package | Before | After |
|---------|--------|-------|
| `env_logger` | 0.10 | 0.11.8 |
| `which` | 6.0 | 8.0.0 |
| `tokio` | 1.0 | 1.49.0 |

### Configuration

- Changed from hardcoded feature list to `"all"` for broader compatibility with feature-gated code

### Testing

All tools have been smoke-tested against a real Rust workspace with:
- Feature-gated modules
- Complex trait hierarchies
- Cross-crate symbol resolution

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -1,8 +1,8 @@
 use anyhow::{anyhow, Result};
-use log::info;
+use log::{info, warn};
 use serde_json::{json, Value};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, VecDeque},
     path::PathBuf,
     process::Stdio,
     sync::Arc,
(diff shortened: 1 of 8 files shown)


src/lsp/client.rs
```diff
@@ -183,24 +272,41 @@ impl RustAnalyzerClient {
...
             return Err(anyhow!("No stdin available"));
         };
 
         stdin.write_all(message.as_bytes()).await?;
         stdin.flush().await?;
 
-        // Set up response channel.
         let (tx, rx) = oneshot::channel();
-        self.pending_requests.lock().await.insert(id, tx);
+        self.pending_requests.lock().await.insert(
```
zeenix, line 283: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] The pending entry is still inserted only after `write_all` and `flush` above. A fast rust-analyzer response can therefore reach the reader first, find no entry, and be discarded, leaving this request to time out. Please base this lifecycle work on #20’s pending-before-write fix (with its cleanup/error-path corrections) rather than retaining this race.

src/lsp/connection.rs
```diff
@@ -115,7 +166,7 @@ async fn handle_lsp_message(
 
     // Check if it's a notification (has method but no id).
     if json_value.get("method").is_some() && json_value.get("id").is_none() {
-        handle_notification(json_value, diagnostics).await;
+        handle_notification(json_value, diagnostics, diagnostics_notify).await;
```
zeenix, line 169: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] This still handles only messages with `method` and no `id`. A rust-analyzer-initiated request such as `workspace/configuration` has both, falls through to response parsing, is ignored or can consume a colliding pending id, and never receives a reply. Please use the corrected server-request transport from #20 and return method-appropriate results.

src/lsp/client.rs
```diff
@@ -138,6 +158,64 @@ impl RustAnalyzerClient {
...
+                    let _ = self.reopen_document(&uri, &content).await;
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Check if rust-analyzer has crashed and attempt restart.
+    pub async fn ensure_healthy(&mut self) -> Result<()> {
```
zeenix, line 212: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Health is checked only from `send_request`, but file tools call `open_document` first and that sends `textDocument/didOpen` through `send_notification`. After rust-analyzer crashes, the first call for any not-yet-open file therefore writes to dead stdin and returns before reaching this restart path; repeated attempts do the same. Check/recover health before notifications or at the workspace/client entry point.

src/lsp/client.rs
```diff
@@ -315,44 +457,170 @@ impl RustAnalyzerClient {
...
+    /// Wait for diagnostics notification with timeout.
+    pub async fn wait_for_diagnostics(&self, uri: &str) -> Option<Vec<Value>> {
+        let timeout = Duration::from_millis(DIAGNOSTICS_WAIT_MILLIS);
+        let start = std::time::Instant::now();
+
+        loop {
+            // Check if we have diagnostics.
+            {
+                let diag_lock = self.diagnostics.lock().await;
+                if let Some(diags) = diag_lock.get(uri) {
```
zeenix, line 583: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Presence is not freshness: rust-analyzer commonly publishes an initial empty diagnostic set after `didOpen` and a later non-empty set after checking. Returning the first cached value—even an empty one—makes `handle_diagnostics` skip its pull fallback and miss real errors. Track a per-URI publication generation captured before `didOpen`, or otherwise wait for the relevant fresh/stable publication; add a regression with empty-then-error notifications.

zeenix (commented):

There is useful tool work here, but this cannot be reviewed or landed as one 886-line commit combining new APIs, diagnostics, restart, request cancellation, document eviction, and rust-analyzer configuration—especially with no tests. The commit subject also lacks the emoji/package prefix required by `CONTRIBUTING.md`, and `git diff --check` reports trailing whitespace at `src/mcp/handlers.rs` lines 210, 245, 464, and 510. Please split this into atomic tested commits after rebasing on the smaller #19 URI and #20 LSP-lifecycle work, preserve current `main`/#23’s shutdown, CLI, and version fixes, and absorb the blockers below into their originating commits.

PR #14, opened 2026-02-23
Fix/skip eager client start on set workspace

abigagli:
When set_workspace was called as the first tool invocation,  ensure_client_started() would spawn rust-analyzer with the default workspace root (often '/') before set_workspace had a chance to update the path.  
This caused rust-analyzer to attempt indexing the entire filesystem, leading to inevitable timeouts on large projects.  
Skip the eager client start for set_workspace since it manages the client lifecycle itself (shutdown old client, update path, start new)

### DIFF
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -48,7 +48,10 @@ pub async fn handle_tool_call(
     tool_name: &str,
     args: Value,
 ) -> Result<ToolResult> {
-    server.ensure_client_started().await?;
+    // set_workspace manages the client lifecycle itself, so skip eager client start.
+    if tool_name != "rust_analyzer_set_workspace" {
+        server.ensure_client_started().await?;
+    }
 
     match tool_name {
         "rust_analyzer_hover" => handle_hover(server, args).await,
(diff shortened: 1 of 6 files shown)


src/mcp/handlers.rs
```diff
@@ -48,7 +48,10 @@ pub async fn handle_tool_call(
     tool_name: &str,
     args: Value,
 ) -> Result<ToolResult> {
-    server.ensure_client_started().await?;
+    // set_workspace manages the client lifecycle itself, so skip eager client start.
+    if tool_name != "rust_analyzer_set_workspace" {
```
zeenix, line 52: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Please add a regression test that makes `rust_analyzer_set_workspace` the very first tool call while the server starts from an unrelated/default cwd. The existing `test_workspace_change` uses a daemon already started and ready with an explicit workspace, so it passes before this fix.

zeenix (commented):

The unique four-line fix is correct, but this branch also contains the first three commits from #9; its head is consequently an ancestor of #9. Please rebase onto current `main` and retain only commit `5263334`. Please also amend its subject to follow `CONTRIBUTING.md` (for example, `🐛 mcp: Skip eager client start for set_workspace`) and add the regression test below. #20 independently contains the same fix, so whichever PR lands later must drop its copy.

PR #15, opened 2026-04-10
fix(mcp): ignore JSON-RPC notifications without id (so can working with codex)

Keinns:
## Summary
  This change makes the MCP server ignore JSON-RPC notifications (requests without `id`) instead of replying with an
  error.

  ## Why
  Some MCP clients send notification messages such as `notifications/initialized` during handshake.
  Per JSON-RPC semantics, notifications must not receive responses. Returning an error response can cause client
  transport/session startup failures.

  ## Change
  - In the MCP request loop, skip response writing when `request.id.is_none()`.
  - Keep existing request/response behavior unchanged for normal requests with `id`.

  ## Validation
  - Built and ran the patched server.
  - Verified `initialize` + `tools/list` still work.
  - Verified notification messages no longer generate error responses.
  - Tested in codex

PR #17, opened 2026-04-20
fix: make rust-analyzer-mcp usable on Windows Codex environments

Keinns:
## Background

  On Windows + Codex CLI, `rust-analyzer-mcp` could list tools, but `tools/call` frequently timed out.

  From investigation, the failures were mainly caused by environment-constrained process startup:

  1. `rust-analyzer` discovery relied too heavily on runtime env vars (`PATH`, `HOME`, `CARGO_HOME`, `RUSTUP_HOME`),
  which may be sanitized in Codex environments.
  2. Even when `rust-analyzer.exe` was found, missing `USERPROFILE`/`HOME` in the child process could cause
  initialization to hang and eventually timeout.

  ## What this PR changes

  ### 1) Stronger rust-analyzer discovery on Windows

  After explicit override (`RUST_ANALYZER_MCP_RUST_ANALYZER_PATH`) and PATH checks, add layered fallbacks:

  - `CARGO_HOME`, `USERPROFILE\\.cargo\\bin`, `HOMEDRIVE+HOMEPATH`
  - `RUSTUP_HOME\\toolchains\\*\\bin`
  - Registry-backed homes from:
    - `HKCU\\Environment`
    - `HKCU\\Volatile Environment`
  - `rustup which rust-analyzer` fallback resolution

  ### 2) Child-process env recovery for rust-analyzer startup

  Before spawning `rust-analyzer` on Windows, if missing in current env, recover and inject:

  - `USERPROFILE`
  - `HOME`
  - `CARGO_HOME`
  - `RUSTUP_HOME`

  This prevents the “binary found, but initialize hangs” failure mode.

  ### 3) Better timeout diagnostics

  LSP timeout errors now include rust-analyzer process status (exited/running), making it easier to distinguish startup
  failures from protocol issues.

  ### 4) Stability improvements in the same execution path

  - LSP request flow now registers `pending` before sending bytes, reducing fast-response race risk.
  - Before first tools call, workspace root can be adjusted from `file_path` (walk up to nearest `Cargo.toml`), avoiding
  startup in a wrong cwd.

  ## Files changed

  - `Cargo.toml` (adds windows-only dependency: `winreg`)
  - `Cargo.lock`
  - `src/lsp/client.rs`
  - `src/mcp/server.rs`
  - `src/mcp/handlers.rs`

  ## Validation

  Validated locally on Windows with MCP `Content-Length` transport:

  - Normal environment: `initialize -> tools/call` succeeds.
  - Constrained environment (HOME/USERPROFILE/CARGO_HOME/RUSTUP_HOME removed, Rust paths removed): `tools/call` still
  succeeds.
  - Release build passes.
----------------
all this is done by codex cli itself, now working on my windows 10 codex cli， maybe it is not the best solution ,just a suggestion.

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -1,10 +1,11 @@
 use anyhow::{anyhow, Result};
-use log::info;
+use log::{info, warn};
 use serde_json::{json, Value};
 use std::{
     collections::{HashMap, HashSet},
-    path::PathBuf,
-    process::Stdio,
+    fs,
+    path::{Path, PathBuf},
+    process::{Command as StdCommand, Stdio},
     sync::Arc,
     time::Duration,
 };
@@ -15,9 +16,33 @@ use tokio::{
 };
 
 use crate::{
-    config::{DOCUMENT_OPEN_DELAY_MILLIS, LSP_REQUEST_TIMEOUT_SECS},
+    config::{lsp_request_timeout_secs, DOCUMENT_OPEN_DELAY_MILLIS, RUST_ANALYZER_PATH_ENV},
     protocol::lsp::LSPRequest,
 };
+use url::Url;
+
+fn to_file_uri(path: &Path) -> Result<String> {
+    #[cfg(windows)]
+    let normalized = {
+        let path_str = path.to_string_lossy();
+        if let Some(stripped) = path_str.strip_prefix(r"\\?\") {
+            PathBuf::from(stripped)
+        } else {
+            path.to_path_buf()
+        }
+    };
+    #[cfg(not(windows))]
+    let normalized = path.to_path_buf();
+
+    Url::from_file_path(&normalized)
+        .map_err(|_| {
+            anyhow!(
+                "Failed to convert path to file URI: {}",
+                normalized.display()
+            )
+        })
+        .map(|u| u.to_string())
+}
 
 pub struct RustAnalyzerClient {
     pub(super) process: Option<Child>,
@@ -73,6 +98,36 @@ impl RustAnalyzerClient {
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
             .stderr(Stdio::piped());
+
+        #[cfg(windows)]
+        {
+            // Windows 下某些 MCP 客户端会裁剪 HOME/USERPROFILE，导致 rust-analyzer 初始化卡死。
+            let profile_for_child = std::env::var("USERPROFILE")
+                .ok()
+                .filter(|v| !v.trim().is_empty())
+                .or_else(resolve_windows_user_profile_from_registry);
+
+            if std::env::var("USERPROFILE").is_err() {
+                if let Some(profile) = &profile_for_child {
+                    cmd.env("USERPROFILE", profile);
+                }
+            }
+            if std::env::var("HOME").is_err() {
+                if let Some(profile) = &profile_for_child {
+                    cmd.env("HOME", profile);
+                }
+            }
+            if std::env::var("CARGO_HOME").is_err() {
+                if let Some(cargo_home) = collect_registry_cargo_homes().into_iter().next() {
+                    cmd.env("CARGO_HOME", cargo_home);
+                }
+            }
+            if std::env::var("RUSTUP_HOME").is_err() {
+                if let Some(rustup_home) = collect_registry_rustup_homes().into_iter().next() {
+                    cmd.env("RUSTUP_HOME", rustup_home);
+                }
+            }
+        }
 
         // Pass through isolation environment variables if they're set.
         if let Ok(cache_home) = std::env::var("XDG_CACHE_HOME") {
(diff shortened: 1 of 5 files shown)


src/lsp/client.rs
```diff
@@ -15,9 +16,33 @@ use tokio::{
 };
 
 use crate::{
-    config::{DOCUMENT_OPEN_DELAY_MILLIS, LSP_REQUEST_TIMEOUT_SECS},
+    config::{lsp_request_timeout_secs, DOCUMENT_OPEN_DELAY_MILLIS, RUST_ANALYZER_PATH_ENV},
```
zeenix, line 19: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] This branch does not compile: `src/config.rs` defines none of `lsp_request_timeout_secs`, `RUST_ANALYZER_PATH_ENV`, or `WORKSPACE_ROOT_ENV` (the latter is also imported by `src/mcp/server.rs`). Please include the intended config changes and run the `CONTRIBUTING.md` build/test/clippy commands before the next review.

src/mcp/handlers.rs
```diff
@@ -48,7 +48,12 @@ pub async fn handle_tool_call(
     tool_name: &str,
     args: Value,
 ) -> Result<ToolResult> {
-    server.ensure_client_started().await?;
+    // `set_workspace` should be able to switch root before the first RA startup.
+    // Otherwise the initial startup may happen in an unrelated large cwd and timeout.
+    if tool_name != "rust_analyzer_set_workspace" {
+        server.adjust_workspace_from_file_arg(&args);
```
zeenix, line 54: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] This can immediately undo an explicit `rust_analyzer_set_workspace`: because that call now starts lazily, the next file tool sees no client and narrows `/monorepo` to the nearest nested crate’s `Cargo.toml`. Please track whether the root was explicitly configured and never auto-replace it; inference should only apply to an implicit default root.

src/lsp/client.rs
```diff
@@ -74,6 +99,36 @@ impl RustAnalyzerClient {
...
 
+        #[cfg(windows)]
+        {
+            // Windows 下某些 MCP 客户端会裁剪 HOME/USERPROFILE，导致 rust-analyzer 初始化卡死。
+            let profile_for_child = std::env::var("USERPROFILE")
+                .ok()
+                .filter(|v| !v.trim().is_empty())
+                .or_else(resolve_windows_user_profile_from_registry);
+
+            if std::env::var("USERPROFILE").is_err() {
```
zeenix, line 110: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: An empty-but-present value is functionally missing here, but `is_err()` leaves it unchanged even when `profile_for_child` recovered a usable registry value. Please use a shared missing-or-empty check for `USERPROFILE`, `HOME`, `CARGO_HOME`, and `RUSTUP_HOME`.

src/lsp/client.rs
```diff
@@ -360,20 +455,288 @@ impl RustAnalyzerClient {
...
+            candidate.display()
+        );
+    }
+
+    if let Ok(found) = which::which("rust-analyzer") {
+        return Ok(found);
+    }
+
+    let candidates = collect_rust_analyzer_candidates();
+    for candidate in &candidates {
```
zeenix, line 475: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This can choose an arbitrary raw toolchain binary because `read_dir` order is unspecified, and it does so before asking rustup which toolchain is active. Please prefer the cargo/rustup proxy or `rustup which rust-analyzer` (run with the workspace as cwd and the recovered homes); do not select directly from `toolchains/*` nondeterministically.

src/lsp/client.rs
```diff
@@ -15,9 +16,33 @@ use tokio::{
...
+    config::{lsp_request_timeout_secs, DOCUMENT_OPEN_DELAY_MILLIS, RUST_ANALYZER_PATH_ENV},
     protocol::lsp::LSPRequest,
 };
+use url::Url;
+
+fn to_file_uri(path: &Path) -> Result<String> {
+    #[cfg(windows)]
+    let normalized = {
+        let path_str = path.to_string_lossy();
+        if let Some(stripped) = path_str.strip_prefix(r"\\?\") {
```
zeenix, line 28: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This normalization breaks verbatim UNC paths: `\\?\UNC\server\share` becomes the relative path `UNC\server\share`. Please remove both duplicated helpers and reuse the centralized path/URI work from #19 after it is fixed and rebased; `url` 2.5.7 handles the verbatim prefixes directly.

src/lsp/client.rs
```diff
@@ -360,20 +455,288 @@ impl RustAnalyzerClient {
...
-        anyhow!(
-            "Failed to find rust-analyzer in PATH or ~/.cargo/bin: {}. Please ensure rust-analyzer is installed.",
-            e
-        )
-    })
+
+        let output = StdCommand::new(&rustup_binary)
+            .args(["which", "rust-analyzer"])
+            .output()
+            .ok()?;
```
zeenix, line 617: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: A spawn failure here returns `None` from the whole function, skipping every remaining rustup candidate. Please log and continue with the next candidate instead of using `?` inside this loop.

zeenix (commented):

This needs to be rebuilt as an atomic Windows-specific change. The branch includes #15 verbatim and duplicates/conflicts with #14, #19, #20, #18, and current `main`, while the main 695-line commit has no detailed body or tests and does not compile. Please rebase after the smaller notification, workspace, URI, and LSP-lifecycle prerequisites; retain only Windows discovery/environment recovery, split any genuinely independent transport work, and use the commit format required by `CONTRIBUTING.md`.

PR #8, opened 2025-11-10
Refactor run() to accept generic async streams

nikomatsakis:
- Add run_with_streams<R, W>() method accepting generic AsyncRead/AsyncWrite streams
- Keep existing run() method for backward compatibility, delegating to run_with_streams()
- Enables library usage with custom streams (duplex, Unix sockets, etc.) beyond stdio

PR #18, opened 2026-05-12
Better mcp series

the-real-wizard-of-oz:
# `better-mcp-series` → `main`

Expands the MCP server from a 10-tool LSP shim into a 36-tool concurrent
service with cancellation, auto-restart, multi-workspace support, and
LLM-friendly output shaping.

**Scope:** 36 commits, +8,694 / −827 lines, 32 files.
**Tests:** 142 green (was ~30) — 89 unit + 28 integration + 7 property + 6 stress + 12 protocol.

## Why

- **Token cost** — raw LSP responses are verbose (a single `documentSymbol`
  could be 60+ KB; `workspace/symbol` returned duplicates; hover markdown
  was uncapped).
- **Round-trip latency** — locating a symbol took 4–5 sequential MCP calls.
- **Operational fragility** — rust-analyzer crashes wedged the server,
  cancellation couldn't reach in-flight LSP work, external edits silently
  desynced rust-analyzer's view, and the post-`didOpen` 200 ms sleep was
  a guess.

## What's new

**26 new tools.** Highlights:

- *Composites* (one round-trip, server-side fan-out via `tokio::join!`):
  `explore_symbol` (hover + def + type_def + parent + refs sample),
  `impact` (refs + callers + implementors + tests — blast radius),
  `get_type_by_name` (no file/position needed).
- *Missing LSP standards*: `rename` + `prepare_rename`, `signature_help`,
  `inlay_hints`, `workspace_symbol`, `type_definition`, `implementation`,
  call/type hierarchy (both directions).
- *rust-analyzer extensions*: `expand_macro`, `parent_module`, `runnables`
  (reshaped envelope) + `run_runnable` (opt-in `cargo` spawn), `related_tests`,
  `open_docs`, `syntax_tree`, `view_hir`, `view_mir`.
- *Refactor*: `move_file` (uses `willRenameFiles` + applies the edit + renames
  on disk).
- *Workspace mgmt*: `add_workspace`, `remove_workspace`, `list_workspaces`.

**MCP resources (new protocol surface).** `resources/list` + `resources/read`
with three URI families: `workspace://files` (tree, capped), `workspace://crates`
(reshaped cargo metadata), `workspace://crate/<name>/Cargo.toml`. Multi-workspace
variants are id-prefixed (`workspace://ws-2/...`); the default workspace keeps
emitting unprefixed legacy URIs.

## Reliability & concurrency

- **True parallelism**: `&mut self` → `Arc<Self>` + spawn-per-request.
- **End-to-end cancellation**: `notifications/cancelled` aborts the spawned
  task *and* forwards `$/cancelRequest` to rust-analyzer. `PendingTracker`
  cross-indexes by LSP id and originating MCP id (via `task_local`), so handler
  signatures stay unchanged.
- **Auto-restart on rust-analyzer crash**: a monitor task drains pending
  oneshots with `LspError::ProcessDied`, next request gets a fresh subprocess.
  Rate-limited (3 crashes / 60 s).
- **External-edit sync**: per-uri version + mtime + content hash;
  `textDocument/didChange` on real edits instead of stale `didOpen` state.
- **`$/progress` integration**: replaced the 200 ms post-`didOpen` sleep with
  a level-triggered watch on rust-analyzer's `rustAnalyzer/cachePriming` token.
- **Typed `LspError`** distinguishes transport / timeout / cancel / process-died
  from "no result at this position" (still surfaced as `null`).

## LLM-friendliness

- **Truncation defaults**: hover ≤ 5 KB, completion ≤ 50 items, `symbols`
  collapses children (with `child_count`), workspace_symbol/diagnostics dedup
  + paginate. `verbose=true` bypasses caps.
- **Pagination**: opaque `cursor` / `next_cursor` on `symbols`,
  `workspace_symbol`, `workspace_diagnostics`.
- **Snippet enrichment**: 9 location-emitting tools auto-attach source
  snippets next to every `{ uri, range }`, eliminating the follow-up `Read`.
  Per-call file cache, 50-hit budget, pagination runs *before* enrichment.
- **`workspace_symbol` dedup** on `(name, containerName, location)` —
  rust-analyzer often returns each entry twice.
- **Diagnostics passthrough**: `data.rendered` (cargo-formatted error block),
  `codeDescription.href`, and `tags` now reach the LLM (previously stripped).

## Multi-workspace + persistence

- `WorkspaceRegistry` owns one entry per registered root, each with its own
  rust-analyzer subprocess, restart history, open-doc cache, and stable id
  (`ws-1`, `ws-2`, ...).
- Roots persist to `workspaces.json` under XDG state dir
  (`RUST_ANALYZER_MCP_STATE_DIR` → `XDG_STATE_HOME` → `$HOME/.local/state`).
  Best-effort; set the env var to `""` to disable. Workspace ids are *not*
  persisted (reissued each boot).

## Operational

Audit of 12 days of MCP-client logs (67 sessions, 17 workspaces) showed every
shutdown hitting the SIGTERM fallback. The final commit fixes that:

- Default tracing filter `info` → `warn` (routine startup logs no longer
  pollute the MCP client's stderr-as-error channel).
- Workspace shutdown parallelized via `join_all` with an 80 ms budget;
  `kill_on_drop` is the backstop.
- `std::process::exit(0)` after `run()` returns — `tokio::io::stdin()`'s
  blocking-pool thread sits in `read(2)` on a pipe the parent hasn't closed,
  which deadlocks `Runtime::drop`. Shutdown latency: **>1,000 ms → ~7–8 ms.**

## Backward compatibility

- All 10 pre-existing tools keep their schemas. Response shapes for `symbols`,
  `workspace_symbol`, `workspace_diagnostics` are now wrapper objects
  (`{ symbols, total, ... }`) — additive, not breaking semantics, but callers
  that destructure the raw array need a one-line adjustment.
- `set_workspace` keeps working (now replaces the default workspace's root
  in place, preserving its id).
- Unprefixed `workspace://...` URIs still route to the default workspace.
- `workspace_id` is optional on every workspace-scoped tool.

## Known limitations / follow-ups

- LSP `$/cancelRequest` arrives a few ms after the local task abort
  (acceptable in practice).
- `get_type_by_name` only resolves what rust-analyzer's `workspace/symbol`
  indexes — typically workspace members; external-crate paths return
  `primary: null`.
- `update_document` (write-through edits) intentionally out of scope; external
  edits via the host's `Edit` tool remain the canonical path.


zeenix (commented):

This special-care pass found several independent merge blockers in cancellation/transport handling, mutating tools, URI/position correctness, and resource bounds. Please do not merge the current head. The safest sequence is to land the smaller prerequisites first (#15 notification handling, #14 workspace startup, corrected #19 URI conversion, corrected #20 LSP lifecycle, a ping-only #9, and #8’s stream abstraction), then rebase this series and drop/port the overlapping implementations. The rebase must preserve all of current `main`/#23: cross-platform `ShutdownSignal`, explicit runtime shutdown, CLI parsing, package-derived server version, shared IPC fixes, and shutdown tests.

Given the 9k-line scope, I strongly recommend splitting this into staged PRs—runtime/lifecycle; read-only tools and shaping; multi-workspace/resources; and the security-sensitive `move_file`/`run_runnable` tools—rather than landing all 40 commits together. `CONTRIBUTING.md` also requires consistent package/component prefixes; commit `46b38cb` is not atomic, and follow-up fixes/tests should be absorbed into their originating commits rather than appended. Finally, the public API and array-to-object response changes are breaking, not additive, so a `0.2.1` patch release is not appropriate unless compatibility is preserved. GitHub Actions currently shows `action_required`, so there is no green CI result; after rebasing, please run the full `cargo test --all-features`, nightly fmt, and clippy checks.

src/mcp/server.rs
```diff
@@ -25,169 +43,468 @@ impl Default for RustAnalyzerMCPServer {
...
+
+        // Forward `$/cancelRequest` to rust-analyzer for any LSP calls this
+        // MCP request had in flight, so the upstream work actually stops
+        // instead of being abandoned. Workspaces only know about their own LSP
+        // ids — broadcast across all of them; clients without matching
+        // entries no-op cheaply. Do this before aborting the spawn so the
+        // tracker still has the LSP ids registered.
+        let entries = self.workspaces.read().await.list();
+        for entry in entries {
+            if let Some(client) = entry.maybe_client().await {
```
zeenix, line 419: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Cancellation is blocked before it reaches `handle.abort()`. A request cold-starting rust-analyzer holds `WorkspaceEntry.client`’s write lock across `client.start().await`; cancelling that request waits here for `maybe_client()`’s read lock, so cancellation cannot take effect until initialization/reload has already finished, potentially after multiple LSP timeouts. Abort promptly before awaiting workspace locks, or make upstream cancellation forwarding non-blocking, and add a cold-start cancellation test.

src/mcp/server.rs
```diff
@@ -25,169 +43,468 @@ impl Default for RustAnalyzerMCPServer {
...
+
+        let response_json = match serde_json::to_string(&response) {
+            Ok(s) => s,
+            Err(e) => {
+                error!("Failed to serialize response: {}", e);
+                return;
+            }
+        };
+
+        let mut writer = stdout.lock().await;
```
zeenix, line 458: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] This request remains registered in `in_flight` while writing its response. A concurrent cancellation can abort `write_all` after only a prefix has reached the pipe, or between the JSON and newline/flush; the next response then appends to an unterminated frame and corrupts stdio for every later request. Remove the request from cancellation tracking before response emission, or enqueue complete frames to a dedicated non-abortable writer. Fatal writer errors must also be sent back to the main loop so cleanup runs.

src/lsp/client.rs
```diff
@@ -180,30 +236,85 @@ impl RustAnalyzerClient {
...
-        stdin.flush().await?;
+        debug!("Sending LSP request: {} with params: {:?}", method, params);
 
-        // Set up response channel.
+        // Register the pending request *before* writing, so a fast response can't race past us.
+        // Inherit the current MCP request id (if any) so cancel_mcp can find this LSP call later.
+        let mcp_request_id = CURRENT_MCP_REQUEST_ID.try_with(|s| s.clone()).ok();
         let (tx, rx) = oneshot::channel();
-        self.pending_requests.lock().await.insert(id, tx);
+        self.pending_requests.insert(
```
zeenix, line 249: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] This pending entry has no drop guard. The composite handlers wrap calls in 2-second `tokio::time::timeout`s (for example `handlers.rs:404` and 840–849), so expiry drops `send_request` before its own timeout reaches the cleanup at line 280. If rust-analyzer never answers, both pending indexes remain forever and no `$/cancelRequest` is sent. Make registration RAII/cancellation-safe and test that dropping a request future removes both indexes and cancels upstream.

src/lsp/connection.rs
```diff
@@ -171,5 +317,226 @@ async fn handle_notification(
...
+    debug!("Server request: {} (id={})", method, id);
+
+    // Reply with `null` to acknowledge. `window/workDoneProgress/create` is the
+    // common case — rust-analyzer may also send `client/registerCapability` etc.
+    // We don't actually register anything, but a successful empty reply keeps
+    // the server happy.
+    let response = json!({
+        "jsonrpc": "2.0",
+        "id": id,
+        "result": null,
```
zeenix, line 435: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] `workspace/configuration` cannot be acknowledged with `null`: its result must be an array with one entry per `params.items`. Returning null can make rust-analyzer reject configuration or continue with unusable settings. Dispatch server-initiated requests by method—#20 already contains the item-count implementation—and return `-32601` for unsupported methods rather than falsely acknowledging them.

src/mcp/server.rs
```diff
@@ -25,169 +43,468 @@ impl Default for RustAnalyzerMCPServer {
...
-        let running = Arc::new(Mutex::new(true));
-        let running_clone = Arc::clone(&running);
+        let stdout = Arc::new(Mutex::new(BufWriter::new(tokio::io::stdout())));
 
-        tokio::spawn(async move {
-            let _ = tokio::signal::ctrl_c().await;
-            info!("Received shutdown signal");
-            *running_clone.lock().await = false;
-        });
+        let mut shutdown = std::pin::pin!(tokio::signal::ctrl_c());
```
zeenix, line 190: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] This branch predates current `main`/#23 and replaces its persistent cross-platform `ShutdownSignal` with one `ctrl_c` future. Keeping this during rebase would again skip SIGTERM/SIGHUP and Windows console-close cleanup, ignore signal-registration errors, and discard the new shutdown coverage. Please adapt the concurrent loop around current main’s signal abstraction and preserve `tests/integration/shutdown.rs`.

src/mcp/server.rs
```diff
@@ -25,169 +43,468 @@ impl Default for RustAnalyzerMCPServer {
...
+                            let _ = handle.await;
+                            in_flight_cleanup.lock().remove(&key);
+                        });
+                    }
+                }
+            }
         }
 
-        // Cleanup.
+        // Cleanup: shutdown every workspace's client. Parallel + tight budget
```
zeenix, line 261: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Shutdown never aborts or joins the spawned request tasks. They may retain clients and keep issuing LSP calls while these shutdown futures run; after 80 ms the code relies on `Drop` even though those `Arc`s are still live. Abort and drain `in_flight` before client teardown, then explicitly force-kill each rust-analyzer on the deadline. Add a signal-during-slow-tool test that verifies no child survives.

src/lsp/connection.rs
```diff
@@ -171,5 +317,226 @@ async fn handle_notification(
...
+    debug!("Progress {} kind={}", token, kind);
+
+    let mut map = progress.lock();
+    let sender = map.entry(token.clone()).or_insert_with(|| {
+        let (tx, _) = watch::channel(false);
+        tx
+    });
+
+    match kind {
+        "begin" | "report" => {
```
zeenix, line 353: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: A new `begin` must reset this token to false. rust-analyzer reuses the literal `rustAnalyzer/cachePriming` token for later priming sequences, so once the first sequence reaches `end`, this leaves the watch true forever and later document opens return immediately during active priming. Handle `begin` and `report` separately and test `begin → end → begin`; a generation counter would also prevent an old `end` from releasing a restarted sequence.

src/mcp/handlers.rs
```diff
@@ -1,328 +1,1368 @@
...
+    let total_budget = tokio::time::Duration::from_secs(8);
+    let pulse_timeout = tokio::time::Duration::from_secs(2);
+    let deadline = tokio::time::Instant::now() + total_budget;
+    while result.as_array().is_some_and(|a| a.is_empty()) {
+        let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
+        if remaining.is_zero() {
+            break;
+        }
+        let wait = pulse_timeout.min(remaining);
+        if client.wait_for_diagnostics_change(wait).await.is_err() {
```
zeenix, line 1209: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: `Notify::notify_waiters` does not retain a permit. A diagnostics publish can land after the initial map read but before this waiter is registered, leaving non-empty diagnostics in the map while this call times out. This path then breaks after only 2 seconds despite the 8-second budget and does not perform the promised final read. Use a watch/generation subscribed before the map read, or at minimum re-read on timeout and continue to the deadline; add a deterministic lost-wakeup test.

src/lsp/error.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// -32603 (e.g. `Invalid offset`, `no rust file`) on top of the standard
+    /// MethodNotFound and the request-state sentinel codes (-32801..=-32803).
+    ///
+    /// Callers that want to distinguish a real internal bug from a lookup miss
+    /// should rely on the `tracing` logs — `lsp/connection.rs` logs every
+    /// non-zero LSP error at `error!` level before this coercion runs.
+    pub fn is_no_result(&self) -> bool {
+        matches!(
+            self,
+            LspError::MethodNotFound { .. } | LspError::InternalError { .. }
```
zeenix, line 52: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: `from_lsp_error` maps every unrecognized code to `InternalError`, and this predicate then turns all of them—including a genuine JSON-RPC `-32603` rust-analyzer fault—into successful `null`. That makes “no symbol here” indistinguishable from a broken analysis request. Preserve/check the numeric code and coerce only specifically known no-result cases; add a test proving a real internal error propagates.

tests/integration/cancellation.rs
```diff
@@ -0,0 +1,70 @@
+use anyhow::{anyhow, Result};
+use serde_json::json;
+
+use test_support::MCPTestClient;
+
+/// `notifications/cancelled` for an in-flight request must not deadlock or
+/// corrupt the server: subsequent requests must still be served. We don't
+/// assert that any *specific* call gets cancelled (that's race-sensitive on
```
zeenix, line 8: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This test never demonstrates active cancellation: it deliberately accepts either outcome, ignores the read result, and delegates the asserted path to an untracked `/tmp` script. Replace it with a deterministic barrier-backed slow request and an ID-aware reader; assert that the cancelled request is removed/cancelled upstream, emits no partial frame, and the next ID receives a valid response.

src/mcp/handlers.rs
```diff
@@ -1,328 +1,1368 @@
...
+    let from_canon = from_abs.canonicalize().unwrap_or_else(|_| from_abs.clone());
+    if !from_canon.starts_with(&canonical_root) {
+        return Err(anyhow!("Source must be inside the workspace: {}", from));
+    }
+    // For `to_abs` we can't canonicalize (doesn't exist yet); compare its parent.
+    let to_parent = to_abs
+        .parent()
+        .map(|p| p.to_path_buf())
+        .unwrap_or_else(|| to_abs.clone());
+    let to_parent_canon = to_parent
```
zeenix, line 657: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Failed canonicalization is not safe here. With `to="../new-outside/file.rs"` and a nonexistent outside parent, the fallback remains a lexical `root/../new-outside`; it still `starts_with(root)`, then `create_dir_all` and `rename` resolve `..` and escape the workspace. Reject parent/prefix components and validate a normalized destination via its canonical nearest-existing ancestor; add this absent-parent traversal case.

src/mcp/workspace_edit.rs
```diff
@@ -0,0 +1,371 @@
...
+                "Refusing to apply edit outside workspace: {}",
+                op.uri
+            ));
+        }
+
+        let original = std::fs::read_to_string(&canonical_path)
+            .map_err(|e| anyhow!("Failed to read {}: {}", canonical_path.display(), e))?;
+        let updated = apply_workspace_edit_str(&original, &op.edits)?;
+        if updated != original {
+            std::fs::write(&canonical_path, &updated)
```
zeenix, line 60: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Files are committed one by one before later edits are even validated, and `move_file` commits all of these before destination creation/rename. A malformed/outside second edit or failed rename therefore leaves imports partially rewritten even though the tool reports failure. Pre-parse/read/validate every operation, serialize mutations per workspace, stage atomic writes, and rollback on write/rename failure. Please test second-file failure, rename failure, and concurrent moves.

src/mcp/workspace_edit.rs
```diff
@@ -0,0 +1,371 @@
...
+/// byte-offset positions for the ASCII edits we expect from
+/// `willRenameFiles`. Returns `None` if the position points past EOF.
+fn position_to_byte(text: &str, line: u32, character: u32) -> Option<usize> {
+    let mut byte_offset = 0usize;
+    let mut current_line = 0u32;
+    for line_text in text.split_inclusive('\n') {
+        if current_line == line {
+            let trimmed = line_text.strip_suffix('\n').unwrap_or(line_text);
+            let char_byte = trimmed
+                .char_indices()
```
zeenix, line 205: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] The client advertises no `general.positionEncodings`, so LSP positions default to UTF-16; this counts Unicode scalar values instead. An astral character before an ASCII import shifts the edit and can corrupt the wrong text. `unwrap_or(trimmed.len())` also silently clamps columns past EOL. Either negotiate and honor UTF-8 or implement correct UTF-16 conversion, distinguish exact EOL from out-of-range, and add astral/out-of-range tests.

src/mcp/handlers.rs
```diff
@@ -1,328 +1,1368 @@
...
+        .stdin(std::process::Stdio::null())
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped());
+
+    let started_at = std::time::Instant::now();
+    let child = cmd
+        .spawn()
+        .map_err(|e| anyhow!("Failed to spawn cargo: {}", e))?;
+
+    let wait = child.wait_with_output();
```
zeenix, line 772: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] `wait_with_output()` buffers all stdout and stderr in memory before the advertised 5 KiB cap is applied. A noisy test or runnable can therefore exhaust memory during the 60–600 second timeout. Drain both pipes concurrently while retaining only the capped prefix and discarding the remainder; cover this with output far larger than the cap.

src/mcp/handlers.rs
```diff
@@ -1,328 +1,1368 @@
...
+        .unwrap_or(DEFAULT_RUN_TIMEOUT_SECS)
+        .min(MAX_RUN_TIMEOUT_SECS);
+
+    let ws = server.resolve_workspace(&args).await?;
+    let workspace_root = ws.root_clone();
+
+    // The Tokio task running this future is the same one the MCP server
+    // tracks via `in_flight`; aborting it drops the Child and SIGKILLs cargo
+    // because of `kill_on_drop`.
+    let mut cmd = tokio::process::Command::new("cargo");
```
zeenix, line 759: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: `kill_on_drop` terminates only the direct `cargo` PID, not necessarily the test/binary process tree it spawned. Timeout or MCP cancellation can leave user code running after the tool reports it was killed. Start and reap a process group/session on Unix and a Job Object on Windows, then test a runnable that spawns a long-lived descendant.

src/mcp/runnables.rs
```diff
@@ -0,0 +1,345 @@
...
+///
+/// Rules: non-empty; first arg must be in [`ALLOWED_CARGO_SUBCOMMANDS`].
+/// Cargo *flags* before the subcommand (e.g. `+nightly`, `-Z something`) are
+/// rejected — the runnable reshape never emits them and accepting them would
+/// open arbitrary command surface.
+pub fn validate_cargo_args(args: &[String]) -> Result<(), String> {
+    let Some(first) = args.first() else {
+        return Err("cargo_args must be non-empty".into());
+    };
+    if !ALLOWED_CARGO_SUBCOMMANDS.contains(&first.as_str()) {
```
zeenix, line 167: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Validating only argv[0] does not keep execution inside the selected workspace: `--manifest-path /outside/Cargo.toml` runs outside it, and `--config build.rustc-wrapper=...` can execute an arbitrary wrapper. Prefer an opaque server-issued runnable ID or exact cached argv; at minimum reject dangerous config/path flags and require any manifest path to canonicalize under the workspace.

src/mcp/handlers.rs
```diff
@@ -1,328 +1,1368 @@
...
+        .map(|p| p.to_path_buf())
+        .unwrap_or_else(|| to_abs.clone());
+    let to_parent_canon = to_parent
+        .canonicalize()
+        .unwrap_or_else(|_| to_parent.clone());
+    if !to_parent_canon.starts_with(&canonical_root) {
+        return Err(anyhow!("Destination must be inside the workspace: {}", to));
+    }
+
+    let from_uri = format!("file://{}", from_canon.display());
```
zeenix, line 664: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Manual `file://{path.display()}` construction is invalid for Windows drive/UNC paths and for paths containing spaces, `#`, or `%`; the matching manual stripping elsewhere cannot percent-decode either. Rebase on the corrected #19 abstraction, extend it with URI-to-path conversion, and port every new callsite in `lsp/client.rs`, `mcp/workspace.rs`, handlers, snippets, and workspace edits.

src/mcp/resources.rs
```diff
@@ -0,0 +1,519 @@
...
+    if depth >= FILE_TREE_MAX_DEPTH {
+        return json!({
+            "type": "dir",
+            "name": name,
+            "truncated": true,
+            "reason": "max-depth"
+        });
+    }
+
+    let mut entries: Vec<_> = match std::fs::read_dir(path) {
```
zeenix, line 245: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The 5,000-entry limit does not bound this operation: `read_dir(...).collect()` materializes and sorts the entire directory before the limit is checked, while symlink/other/unknown nodes return without incrementing `count` and can exceed the output cap indefinitely. Count every emitted node and use a bounded streaming/selection strategy; test a >cap symlink farm and a very large flat directory.

src/mcp/snippets.rs
```diff
@@ -0,0 +1,462 @@
...
+            return None;
+        }
+
+        let mut total_bytes = 0usize;
+        let mut out: Vec<String> = Vec::with_capacity(to - from);
+        let mut hit_byte_cap = false;
+        for line in &lines[from..to] {
+            // +1 for the implicit '\n' so the cap reflects rendered size
+            let line_bytes = line.len().saturating_add(1);
+            if total_bytes.saturating_add(line_bytes) > SNIPPET_MAX_BYTES_PER_HIT && !out.is_empty()
```
zeenix, line 111: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The `!out.is_empty()` exception admits an arbitrarily long first line whole, so a 1 MiB line produces a 1 MiB “400-byte-capped” snippet. The implementation also reads and splits the entire file synchronously on an async worker before enforcing the output cap. Bound input reads/cache, move blocking I/O off the Tokio worker, and truncate a long individual line at a UTF-8 boundary.

src/mcp/truncate.rs
```diff
@@ -0,0 +1,712 @@
...
+        return value;
+    }
+    let Value::Array(items) = value else {
+        return value;
+    };
+
+    let items = dedup_workspace_symbols(items);
+    let total = items.len();
+    let start = cursor.min(total);
+    let take = if verbose { total - start } else { limit };
```
zeenix, line 290: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: `verbose=true` ignores the already-resolved `limit`, including the schema’s documented 1,000-item absolute cap, and returns every remaining symbol. The same pattern appears for completion and workspace diagnostics. Always take `limit.min(remaining)`; verbose should change the default limit, not bypass the absolute/explicit one. Add >1,000-item and `verbose + explicit limit` tests.

src/mcp/persistence.rs
```diff
@@ -0,0 +1,183 @@
...
+    };
+    let body = serde_json::to_vec_pretty(&on_disk).context("serializing workspaces.json")?;
+    {
+        let mut f = fs::File::create(&tmp)
+            .with_context(|| format!("creating tempfile {}", tmp.display()))?;
+        f.write_all(&body)
+            .with_context(|| format!("writing tempfile {}", tmp.display()))?;
+        f.sync_all().ok();
+    }
+    fs::rename(&tmp, &target)
```
zeenix, line 106: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: `std::fs::rename(tmp, target)` does not replace an existing destination on Windows. The first save succeeds, later registry saves fail and are only logged, so restart restores stale workspace roots. Use a cross-platform atomic-replace strategy and test two consecutive saves/replacements on Windows.

src/main.rs
```diff
@@ -1,22 +1,34 @@
...
+    // `debug`) to opt into verbose logging.
+    fmt()
+        .with_env_filter(
+            EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("warn")),
+        )
+        .with_writer(std::io::stderr)
+        .init();
 
-    // Get workspace path from command line or use current directory.
     let workspace_path = std::env::args()
```
zeenix, line 19: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Current `main` now uses `args_os` with help/version/unknown-option/extra-argument handling and an explicit runtime with `shutdown_background` plus panic handling. This branch restores `args()` (including the non-UTF-8 panic and treating flags as workspace paths) and ends with `process::exit(0)`, which skips destructors. Resolve this conflict in favor of current main’s CLI/runtime structure while adapting the new server call.

Cargo.toml
```diff
@@ -4,7 +4,7 @@ exclude = ["test-project", "test-project-diagnostics"]
 
 [package]
 name = "rust-analyzer-mcp"
-version = "0.2.0"
+version = "0.2.1"
```
zeenix, line 7: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This series changes several public response types from arrays to wrapper objects and changes the public server API, so `0.2.1` is not an appropriate patch bump unless backward compatibility is restored; for a pre-1.0 crate this should be a `0.3.0`-style breaking release. Also preserve current main’s `env!("CARGO_PKG_VERSION")` initialize response—the PR still reports hard-coded `0.1.0` at `src/mcp/server.rs:480`.

PR #19, opened 2026-06-14
🐛 mcp: Generate valid file URIs on Windows

psyche314:
## Summary

- add shared path helpers for absolute path normalization and file/directory URI conversion
- build LSP `rootUri` and opened document URIs with `url::Url`
- normalize Windows verbatim paths before turning them into LSP file URIs

## Why

Formatting `PathBuf::display()` into `file://...` produces invalid file URIs on Windows, for example drive-letter paths are emitted as `file://C:\...`. rust-analyzer rejects those URIs, so MCP file-level requests can fail before analysis starts.

## Validation

- `cargo +nightly fmt --all`
- `cargo clippy -- -D warnings`
- `cargo build --release`

`cargo test --all-features` and `cargo test -p rust-analyzer-mcp --lib` currently do not compile on Windows before reaching this change because `test-support` imports `std::os::unix::net::{UnixStream, UnixListener}`.

### DIFF
--- a/src/paths.rs
+++ b/src/paths.rs
@@ -1,0 +1,50 @@
+use anyhow::{anyhow, Result};
+use std::path::{Path, PathBuf};
+use url::Url;
+
+pub fn absolute_path(path: PathBuf) -> PathBuf {
+    let path = if path.is_absolute() {
+        path
+    } else {
+        std::env::current_dir()
+            .unwrap_or_else(|_| PathBuf::from("."))
+            .join(path)
+    };
+
+    normalize_path(path)
+}
+
+pub fn file_uri(path: &Path) -> Result<String> {
+    Url::from_file_path(path)
+        .map(|url| url.to_string())
+        .map_err(|_| anyhow!("failed to convert path to file URI: {}", path.display()))
+}
+
+pub fn directory_uri(path: &Path) -> Result<String> {
+    Url::from_directory_path(path)
+        .map(|url| url.to_string())
+        .map_err(|_| {
+            anyhow!(
+                "failed to convert path to directory URI: {}",
+                path.display()
+            )
+        })
+}
+
+#[cfg(windows)]
+fn normalize_path(path: PathBuf) -> PathBuf {
+    let path_str = path.to_string_lossy();
+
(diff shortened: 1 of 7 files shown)


zeenix (commented):

Using `url::Url` is the right isolated direction, and this should become the shared URI foundation before #17 and #18 are rebased. The current draft still has the filesystem-path identity problem and missing Windows regression coverage below, and it needs a small rebase over the import changes in current `main`. #17 should drop its overlapping URI implementation; #18 must port the accepted helpers instead of continuing to format and strip `file://` manually.

src/paths.rs
```diff
@@ -0,0 +1,50 @@
...
+pub fn absolute_path(path: PathBuf) -> PathBuf {
+    let path = if path.is_absolute() {
+        path
+    } else {
+        std::env::current_dir()
+            .unwrap_or_else(|_| PathBuf::from("."))
+            .join(path)
+    };
+
+    normalize_path(path)
```
zeenix, line 14: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Please preserve the previous canonicalize-if-possible behavior and do not strip the resulting Windows verbatim prefix. `url` 2.5.7 already handles `Prefix::VerbatimDisk` and `Prefix::VerbatimUNC` directly. As written, symlinked roots/files lose their canonical identity, while `normalize_path` removes `\\?\` semantics from the `PathBuf` later used by `Command::current_dir` and `read_to_string`—and does so through a lossy Unicode conversion. Please keep the canonical/verbatim filesystem path and pass it directly to `Url`.

src/paths.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+pub fn file_uri(path: &Path) -> Result<String> {
```
zeenix, line 17: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Please add Windows regression tests covering drive-letter paths, UNC paths, verbatim drive/UNC paths, and URI-reserved characters such as spaces, `#`, and `%`. This PR’s build-only Windows validation does not exercise the behavior being fixed; the Unix-only test-support setup should be gated or otherwise adjusted so these unit tests can run on Windows CI.

PR #20, opened 2026-06-14
🐛 lsp: Avoid request response races

psyche314:
## Summary

- register pending LSP response channels before writing requests to rust-analyzer
- share the rust-analyzer stdin writer with the stdout reader so server-side LSP requests can be answered
- respond to `workspace/configuration` requests from rust-analyzer
- cancel pending LSP requests when rust-analyzer stdout closes or errors
- restart the client when the rust-analyzer subprocess has exited
- avoid starting a default client before handling `rust_analyzer_set_workspace`
- keep shutdown best-effort with a short timeout before killing the subprocess
- remove the synchronous `rust-analyzer/reloadWorkspace` request during initialization

## Why

Some rust-analyzer responses can arrive immediately. The previous ordering wrote the LSP request first and only registered the pending response channel afterwards, so a fast response could be dropped and the caller would wait until the request timeout.

The bridge also did not answer rust-analyzer client-side requests such as `workspace/configuration`, which can leave rust-analyzer waiting for a client response. In addition, `rust_analyzer_set_workspace` first started a client for the current/default directory and then immediately shut it down, so workspace switching could spend a full timeout on work that was about to be discarded.

If rust-analyzer exits while requests are pending, the bridge should cancel those requests and allow a later tool call to restart the client instead of keeping a dead client around and timing out repeatedly.

## Validation

- `cargo +nightly fmt --all`
- `git diff --check`
- `cargo clippy -- -D warnings`
- `cargo build --release`

`cargo test --all-features` currently does not compile on Windows before reaching this change because `test-support` imports `std::os::unix::net::{UnixStream, UnixListener}`.

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -23,7 +23,7 @@ pub struct RustAnalyzerClient {
     pub(super) process: Option<Child>,
     pub(super) request_id: Arc<Mutex<u64>>,
     pub(super) workspace_root: PathBuf,
-    pub(super) stdin: Option<BufWriter<tokio::process::ChildStdin>>,
+    pub(super) stdin: Option<Arc<Mutex<BufWriter<tokio::process::ChildStdin>>>>,
     pub(super) pending_requests: Arc<Mutex<HashMap<u64, oneshot::Sender<Value>>>>,
     pub(super) initialized: bool,
     pub(super) open_documents: Arc<Mutex<HashSet<String>>>,
@@ -102,12 +102,14 @@ impl RustAnalyzerClient {
             .take()
             .ok_or_else(|| anyhow!("Failed to get stderr"))?;
 
-        self.stdin = Some(BufWriter::new(stdin));
+        let stdin = Arc::new(Mutex::new(BufWriter::new(stdin)));
+        self.stdin = Some(Arc::clone(&stdin));
 
         // Start connection handlers.
         super::connection::start_handlers(
             stdout,
             stderr,
+            stdin,
             Arc::clone(&self.pending_requests),
             Arc::clone(&self.diagnostics),
         );
@@ -138,6 +140,14 @@ impl RustAnalyzerClient {
         Ok(())
     }
 
+    pub fn is_running(&mut self) -> bool {
+        let Some(process) = &mut self.process else {
+            return false;
+        };
+
+        matches!(process.try_wait(), Ok(None))
+    }
+
     pub(super) async fn send_notification(
         &mut self,
         method: &str,
@@ -154,10 +164,11 @@ impl RustAnalyzerClient {
 
         info!("Sending LSP notification: {}", method);
 
-        let Some(stdin) = &mut self.stdin else {
+        let Some(stdin) = &self.stdin else {
             return Err(anyhow!("No stdin available"));
         };
 
+        let mut stdin = stdin.lock().await;
         stdin.write_all(message.as_bytes()).await?;
         stdin.flush().await?;
         Ok(())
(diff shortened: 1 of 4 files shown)


src/lsp/client.rs
```diff
@@ -102,12 +102,14 @@ impl RustAnalyzerClient {
...
 
-        self.stdin = Some(BufWriter::new(stdin));
+        let stdin = Arc::new(Mutex::new(BufWriter::new(stdin)));
+        self.stdin = Some(Arc::clone(&stdin));
 
         // Start connection handlers.
         super::connection::start_handlers(
             stdout,
             stderr,
+            stdin,
```
zeenix, line 112: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] The detached handler now keeps child stdin alive. If `initialize()` below fails, the caller never stores this client; dropping it neither kills the `Child` nor closes stdin because this `Arc` remains, so rust-analyzer and the handler tasks can leak indefinitely. Please enable `kill_on_drop(true)` before spawning and explicitly kill+wait/abort handlers on every post-spawn error path.

src/lsp/client.rs
```diff
@@ -138,6 +140,14 @@ impl RustAnalyzerClient {
         Ok(())
     }
 
+    pub fn is_running(&mut self) -> bool {
```
zeenix, line 143: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Process liveness is not sufficient client health: the stdout handler can terminate on a read/framing error while the child remains alive. In that state this returns true forever and every later request has no reader. Please have the handler mark the transport dead on all terminal paths and include that in the restart decision.

src/lsp/connection.rs
```diff
@@ -143,6 +165,49 @@ async fn handle_lsp_message(
...
+    let result = match method {
+        "workspace/configuration" => {
+            let item_count = json_value
+                .get("params")
+                .and_then(|params| params.get("items"))
+                .and_then(|items| items.as_array())
+                .map_or(0, Vec::len);
+            json!(vec![json!({}); item_count])
+        }
+        _ => Value::Null,
```
zeenix, line 186: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Returning successful `null` for every unknown server request can falsely acknowledge work we did not perform, and is the wrong result type for requests such as `workspace/applyEdit`. Please explicitly handle supported methods and send JSON-RPC `-32601 Method not found` for the rest.

zeenix (commented):

The pending-before-write fix is sound, but this single commit also bundles server-request support, process restart, shutdown/reload semantics, and #14’s workspace fix without regression tests. Please rebase on current `main`, keep #23’s shutdown implementation, drop the #14 duplication, and split the remaining lifecycle changes into atomic tested `lsp:` commits. #18 independently contains versions of all these changes, so the smaller corrected implementation should either land before #18 and be absorbed there, or this PR should be closed as superseded—both implementations must not merge independently.

src/lsp/client.rs
```diff
@@ -185,22 +196,50 @@ impl RustAnalyzerClient {
 
         info!("Sending LSP request: {} with params: {:?}", method, params);
 
-        let Some(stdin) = &mut self.stdin else {
+        // Register the pending response before writing the request. Some LSP
```
zeenix, line 199: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Please add deterministic regression coverage for the ordering fixed here (a response arriving immediately after the write), plus server-initiated configuration, stdout/process death cancelling pending calls, and restart. This branch currently changes all of those lifecycle paths without adding any tests.

PR #22, opened 2026-08-14
🐛 fix(protocol): silently handle MCP notifications without emitting error responses

nazmulidris:
# Repro Steps

1. Install `rust-analyzer-mcp` using `cargo install --force rust-analyzer-mcp`.

2. [Install](https://antigravity.google/docs/cli/reference) Antigravity CLI (`agy`) and
   configure it to use `rust-analyzer-mcp` as an MCP server named `rust-analyzer` by
   adding the following configuration to your Rust project's `.gemini/settings.json` (or
   global `~/.gemini/config/mcp_config.json`):

   ```json
   {
     "mcpServers": {
       "rust-analyzer": {
         "command": "rust-analyzer-mcp",
         "args": []
       }
     }
   }
   ```

3. Open Antigravity CLI (`agy`) in a folder containing a Rust project and run the `/mcp`
   slash command. You will see the following error:

   ```text
   >  ✗ rust-analyzer  error: failed to get tools: calling "tools/list": invalid request
   ```

> Note: These repro steps also apply to Antigravity IDE.

# Expected Behavior

The [JSON-RPC 2.0][1] and [Model Context Protocol (MCP)][2] specifications state that
notifications (such as `notifications/initialized` and `notifications/cancelled`) do not
contain an `id` and should not write a response or error to `stdout`.

[1]: https://www.jsonrpc.org/specification#notification
[2]: https://modelcontextprotocol.io/docs/concepts/architecture#notifications

Currently, receiving `notifications/initialized` causes `rust-analyzer-mcp` to execute the
wildcard error branch in `handle_request`, returning:

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32601,
    "message": "Method not found: notifications/initialized",
    "data": null
  }
}
```

Because this unexpected error response is written to `stdout`, MCP clients (such as
Google Antigravity CLI / `agy`) that are waiting for the response to their next request
(e.g. `tools/list` with `id: 2`) get out of sync and fail with:

```text
>  ✗ rust-analyzer  error: failed to get tools: calling "tools/list": invalid request
```

# Fix

1. **Silent Notification Handling:** Updated `handle_request()` to return
   `Option<MCPResponse>`. Notifications (where `request.id.is_none()` or
   `request.method.starts_with("notifications/")`) now return `None`.
2. **Output Stream Control:** Updated the `run()` loop to only serialize and write a
   response to `stdout` when `handle_request()` returns `Some(response)`.
3. **Unit Tests:** Added a unit test in `tests/unit/protocol/request_tests.rs` verifying
   notification deserialization without `id`.

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -126,11 +126,14 @@ impl RustAnalyzerMCPServer {
             };
 
             debug!("Received request: {}", request.method);
-            let response = self.handle_request(request).await;
-            let response_json = serde_json::to_string(&response)?;
-            writer.write_all(response_json.as_bytes()).await?;
-            writer.write_all(b"\n").await?;
-            writer.flush().await?;
+            // Only send a response if handle_request returned Some(response) (notifications return None).
+            // See handle_request() for details on how notifications are handled.
+            if let Some(response) = self.handle_request(request).await {
+                let response_json = serde_json::to_string(&response)?;
+                writer.write_all(response_json.as_bytes()).await?;
+                writer.write_all(b"\n").await?;
+                writer.flush().await?;
+            }
         }
 
         // Cleanup.
@@ -142,8 +145,15 @@ impl RustAnalyzerMCPServer {
         Ok(())
     }
 
-    async fn handle_request(&mut self, request: MCPRequest) -> MCPResponse {
-        match request.method.as_str() {
+    async fn handle_request(&mut self, request: MCPRequest) -> Option<MCPResponse> {
+        // Handle MCP notifications (e.g. notifications/initialized, notifications/cancelled).
+        // Per JSON-RPC 2.0 and MCP spec, notifications do not contain an id and MUST NOT produce a response.
+        if request.id.is_none() || request.method.starts_with("notifications/") {
+            debug!("Silently handled notification: {}", request.method);
+            return None;
+        }
+
+        let response = match request.method.as_str() {
             "initialize" => MCPResponse::Success {
                 jsonrpc: "2.0".to_string(),
                 id: request.id,
@@ -167,7 +177,7 @@ impl RustAnalyzerMCPServer {
             },
             "tools/call" => {
                 let Some(params) = request.params else {
-                    return MCPResponse::Error {
+                    return Some(MCPResponse::Error {
                         jsonrpc: "2.0".to_string(),
                         id: request.id,
                         error: MCPError {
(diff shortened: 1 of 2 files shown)


src/mcp/server.rs
```diff
@@ -142,8 +145,15 @@ impl RustAnalyzerMCPServer {
         Ok(())
     }
 
-    async fn handle_request(&mut self, request: MCPRequest) -> MCPResponse {
-        match request.method.as_str() {
+    async fn handle_request(&mut self, request: MCPRequest) -> Option<MCPResponse> {
+        // Handle MCP notifications (e.g. notifications/initialized, notifications/cancelled).
+        // Per JSON-RPC 2.0 and MCP spec, notifications do not contain an id and MUST NOT produce a response.
+        if request.id.is_none() || request.method.starts_with("notifications/") {
```
zeenix, line 151: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: JSON-RPC defines a notification solely by the absence of `id`; the method-name prefix is not part of that classification. An ID-bearing `notifications/...` message is a request and must receive a response (normally an error), whereas this branch silently drops it and can leave the caller waiting forever. Please remove the `starts_with("notifications/")` arm and check only the absence of `id`.

tests/unit/protocol/request_tests.rs
```diff
@@ -103,6 +103,14 @@ fn test_notification_without_id() {
     assert!(deserialized.id.is_none());
 }
 
+#[test]
+fn test_mcp_notifications_initialized() {
```
zeenix, line 107: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This test passes before the fix: it only verifies deserialization into the locally redefined test `MCPRequest`, and the preceding test already covers a missing `id`. Please replace it with an output-level regression test that sends `notifications/initialized`, then an ID-bearing `tools/list`, and asserts that the first response carries the `tools/list` ID. That directly catches the original stdout desynchronization.

src/mcp/server.rs
```diff
@@ -126,11 +126,14 @@ impl RustAnalyzerMCPServer {
             };
 
             debug!("Received request: {}", request.method);
-            let response = self.handle_request(request).await;
-            let response_json = serde_json::to_string(&response)?;
-            writer.write_all(response_json.as_bytes()).await?;
-            writer.write_all(b"\n").await?;
-            writer.flush().await?;
+            // Only send a response if handle_request returned Some(response) (notifications return None).
```
zeenix, line 129: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The current CI run fails formatting here (and on the notification comment below), so the test job never ran. Please run `cargo +nightly fmt --all` and absorb the formatting plus other review fixes into this existing commit, as requested by `CONTRIBUTING.md`.

zeenix (commented):

This has useful reproduction detail, but #15 is the smaller correct duplicate and should be the canonical fix. Independently, the current head has the correctness, test, and formatting blockers called out below. I recommend closing this as superseded once #15 lands; if it is continued instead, please rebase it onto current `main` and amend all fixes into the existing commit as required by `CONTRIBUTING.md`.

PR #23, opened 2026-08-15
🐛 Fix server never exiting on SIGINT/SIGTERM

zeenix:
The server never exited on Ctrl+C or SIGTERM unless its stdin also reached EOF: the signal task only flipped a flag the blocked read loop never re-checked, tokio's installed handler swallowed every follow-up signal, and even with the loop fixed the runtime drop waited forever on the uncancellable blocking stdin read. This left orphaned `rust-analyzer-mcp` processes behind whenever an MCP host signalled without closing the pipe.

This PR makes shutdown signal-driven end to end: persistent SIGINT/SIGTERM/SIGHUP streams (Ctrl+C/console-close on Windows) raced against both the stdin read and in-flight request handling via biased `select!`, a bounded graceful LSP handshake with a force-kill fallback, `shutdown_background()` so the parked stdin read cannot hold the process hostage, and cleanup that runs on stdout errors while still reporting them through the exit status.

Along the way: `--help`/`--version`/argument validation (previously `--help` started a server with workspace `--help`), the initialize response now reports the real crate version, and the dormant shared IPC client tests are revived — ported to `IpcClient` and their daemon fixed to serve clients concurrently instead of deadlocking the second client. New integration tests cover SIGINT/SIGTERM/SIGHUP/stdin-EOF shutdown, including the graceful-handshake path with a live rust-analyzer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0197e2KbPmQpwuFsiiz9iGYa

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -75,6 +75,12 @@ impl RustAnalyzerMCPServer {
         Ok(uri)
     }
 
+    /// Runs the server until its stdin reaches EOF or a shutdown signal arrives.
+    ///
+    /// Installs process-wide signal handlers that remain in effect after this returns. Reads
+    /// stdin through [`tokio::io::stdin`], whose parked blocking read cannot be cancelled: after
+    /// a signal-triggered exit the caller must not wait for the runtime to shut down on its own.
+    /// See this crate's `main.rs`, which uses [`tokio::runtime::Runtime::shutdown_background`].
     pub async fn run(&mut self) -> Result<()> {
         info!("Starting rust-analyzer MCP server");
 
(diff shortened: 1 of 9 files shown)


src/mcp/server.rs
```diff
@@ -126,20 +136,69 @@ impl RustAnalyzerMCPServer {
...
+            let written = async {
+                writer.write_all(response_json.as_bytes()).await?;
+                writer.write_all(b"\n").await?;
+                writer.flush().await
+            };
+            if let Err(e) = written.await {
+                error!("Error writing to stdout: {}", e);
+                result = Err(e.into());
+                break;
+            }
```
zeenix, line 180: 🤖 ChatGPT GPT-5 on behalf of zeenix: The signal is not polled while this write/flush future is running. If the host stops reading stdout and a sufficiently large response fills the pipe, this can remain pending indefinitely and SIGINT/SIGTERM will not terminate the server—the same class of shutdown hang this PR fixes. Please race the complete write operation against `shutdown.recv()` as well.

src/mcp/server.rs
```diff
@@ -86,29 +89,36 @@ impl RustAnalyzerMCPServer {
...
+                    info!("Received shutdown signal");
+                    signals_seen += 1;
                     break;
                 }
+                read = reader.read_line(&mut line) => match read {
+                    Ok(n) => n,
+                    Err(e) => {
+                        error!("Error reading from stdin: {}", e);
+                        break;
+                    }
```
zeenix, line 121: 🤖 ChatGPT GPT-5 on behalf of zeenix: This fatal stdin read error is logged but never stored in `result`, so cleanup runs and the process exits successfully. Please assign `result = Err(e.into())` before breaking, consistently with the stdout error path below.

zeenix (commented):
🤖 ChatGPT GPT-5 on behalf of zeenix: Two blocking shutdown-path issues need addressing before merge. GitHub does not permit requesting changes on your own PR, so this is submitted as a comment review.

src/mcp/server.rs
zeenix: Fixed in 4547074: the response write/flush future is now raced against `shutdown.recv()` in the same biased select as the request handling, so a stdout pipe that stops draining can no longer mask signals.

src/mcp/server.rs
zeenix: Fixed in 4547074: the stdin read error is now stored in `result` before breaking, so after cleanup the process exits non-zero, consistent with the stdout error path.

PR #24, opened 2026-08-15
💚 Fix test_shared_concurrent flake that broke CI on main

zeenix:
Fixes the CI failure on main introduced by #23 (run [31898222002](https://github.com/zeenix/rust-analyzer-mcp/actions/runs/31898222002)).

## What happened

`test_shared_concurrent` — enabled by #23 — failed on the first main run after the merge with:

```
Error: MCP error: {"code":-1,"data":null,"message":"Request timeout"}
```

The same tree (425bb89) had passed CI on the PR branch, so the failure is load-dependent, not
deterministic.

## Root cause

The MCP server bounds every LSP request at 30 seconds (`LSP_REQUEST_TIMEOUT_SECS`). In the failing
run, `test_workspace_diagnostics` (a cold `cargo check` of the whole workspace) ran in parallel on
the ~2-vCPU runner until one second before the failure; one of the five serialized
`documentSymbol` requests exceeded 30 s inside rust-analyzer, the server returned its timeout
error, and the IPC daemon faithfully forwarded it. The rest of the suite tolerates transient
slowness by polling with a deadline (`wait_for_ready`, `initialize_and_wait`, the diagnostics
tests); `IpcClient` was the one path with no tolerance at all.

An adversarial review of the initial fix also surfaced a second, deterministic source of the same
error: `send_request` in the LSP client wrote the request before registering its response channel,
so a reply landing in that window was silently dropped and manufactured a spurious 30-second
timeout.

## Fix

- The LSP client registers the response channel before writing the request, and unregisters it on
  write failure or timeout so abandoned requests cannot leak pending entries.
- `IpcClient::call_tool` retries the server's exact timeout error (typed `McpError`, code and
  message matched — not a substring check), bounded at 3 attempts with backoff via the existing
  `tool_retry_delay()`. Other errors are not retried.
- Since retries put more than one request id on the wire per call, response correlation is now
  enforced at both hops: the daemon refuses to relabel a frame whose id is not the one it just
  forwarded and retires the desynced pipe instead (likewise for unparseable frames), restores the
  client's original id when forwarding a valid response, and the client rejects mismatched ids — a
  desynced shared pipe fails fast instead of returning someone else's answer.
- Unit tests cover both hops: fake-daemon tests for the client (retry uses a fresh id with an
  unchanged payload, bounded give-up, no-retry-on-other-errors, id-mismatch rejection) and
  forwarding-path tests driving `handle_client` against a shell child standing in for the MCP
  server (id restoration, mismatch shutdown, parse-failure shutdown). TDD: the retry test
  reproduced the exact CI error before the fix.
- CI now runs `cargo test --workspace`, since the root-package invocation never ran the
  test-support crate's own unit tests, including the new ones.

Verified locally with the exact CI invocation (`RUSTFLAGS="-D warnings" cargo test --locked
--workspace --release --all-features`): 51 tests across 10 suites pass; fmt and clippy gates clean.

### DIFF
--- a/test-support/src/ipc/server.rs
+++ b/test-support/src/ipc/server.rs
@@ -14,6 +14,10 @@ use std::{
     time::{Duration, Instant},
 };
 
+/// Shared handle to the MCP server's stdio, keeping each request write paired with its
+/// response read.
+type SharedPipes = Arc<Mutex<(ChildStdin, BufReader<ChildStdout>)>>;
+
 /// Start a standalone MCP server process that listens on Unix socket
 pub fn start_server(workspace_path: &Path, project_type: &str) -> Result<()> {
     let socket_path = socket_path(project_type);
@@ -188,7 +192,7 @@ pub fn start_server(workspace_path: &Path, project_type: &str) -> Result<()> {
 /// Serve one client connection until it disconnects.
 fn handle_client(
     mut stream: UnixStream,
-    pipes: Arc<Mutex<(ChildStdin, BufReader<ChildStdout>)>>,
+    pipes: SharedPipes,
     last_activity: Arc<Mutex<Instant>>,
     request_id: Arc<AtomicU64>,
     shutdown: &AtomicBool,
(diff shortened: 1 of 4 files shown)


test-support/src/ipc/server.rs
```diff
@@ -264,8 +264,13 @@ fn handle_client(
             Ok(_) => {}
         }
 
-        // Forward response to client.
-        stream.write_all(response_line.as_bytes())?;
+        // Forward the response to the client with its original request id restored: the daemon
+        // rewrites ids on the shared pipe, and the client correlates responses by the id it
+        // sent.
+        let mut response: Value = serde_json::from_str(&response_line)?;
+        response["id"] = request["id"].clone();
```
zeenix, line 296: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: [blocking] Please validate that the MCP response’s current `id` equals the daemon-generated forwarded `id` before overwriting it with the client’s original id. As written, a stale or unsolicited frame on the shared pipe is relabeled as the current request, and the new client-side check then accepts it—the exact desynchronization this change intends to detect. The mismatch unit test connects directly to the fake daemon, so it bypasses this rewrite and cannot catch the problem. On mismatch, fail the request and retire the shared daemon/pipe; add a test through this forwarding path.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This PR arrived during the final refresh today, so unlike the older queue it was not delayed by the accidentally disabled repository notifications.

The production pending-before-write fix, typed exact retry classification, bounded backoff, and workspace-wide CI command are well structured; the current head is based directly on `main`, merges cleanly, and all three checks are green. There is one response-correlation blocker below. Once corrected, this should land early: it supersedes the pending-ordering slice of #20 (and duplicate versions in #11/#17/#18), while #20’s server-request/process-health work remains separate. Please also add the package/component prefixes requested by `CONTRIBUTING.md` when amending the three subjects (for example `🐛 lsp:`, `💚 test-support:`, and `👷 ci:`).

test-support/src/ipc/server.rs
zeenix: The daemon now validates the frame's id against the forwarded id before relabeling: on mismatch it fails the request and retires itself, consistent with the existing pipe-failure paths (the client sees the dropped connection rather than an error frame, which the retry logic correctly treats as non-retryable). An unparseable frame gets the same treatment, since the pipe is equally unusable after one.

Covered by new tests that drive `handle_client` through the real forwarding path with a shell child standing in for the MCP server: an echo server proving id restoration, a fixed-wrong-id server proving the mismatch shutdown, and a garbage server proving the parse-failure shutdown. The direct-connection mismatch test in client.rs remains as coverage for the client-side check.

Commit subjects now carry the package prefixes (`lsp:`, `test-support:`, `ci:`).

PR #25, opened 2026-08-21
🚑️ Fix rust-analyzer dying on didSave during workspace load, which broke CI on main

zeenix:
Fixes the CI failure on main for the 0.3.0 release (run [32472367064](https://github.com/zeenix/rust-analyzer-mcp/actions/runs/32472367064)).

## What happened

All four `integration::diagnostics::*` tests failed with

```
Error: MCP error -1: Broken pipe (os error 32)
```

That error is `rust-analyzer-mcp` failing to write to rust-analyzer's stdin: rust-analyzer had died
inside the shared `test-project-diagnostics` daemon, and since the server never noticed, every later
tool call failed the same way. The run was the first on the Rust 1.98.0 toolchain (released
2026-08-20); the earlier red runs had been the unrelated "Request timeout" flake fixed in #24.

## Root cause

Reproduced locally with the 1.98.0 rust-analyzer, and — once the daemon kept the MCP server's
stderr — rust-analyzer's own panic showed up:

```
thread 'LspServer' panicked at crates/base-db/src/lib.rs:180:21:
Unable to get `FileSourceRootInput` with `vfs::FileId` (FileId(6), path: <unknown>); this is a bug
  <base_db::Files>::file_source_root → <ide::Analysis>::source_root_id → GlobalState::on_notification
notification: textDocument/didSave
```

`open_document()` sends `textDocument/didSave` right after `didOpen`. While rust-analyzer is still
loading the workspace it only applies VFS changes to its database between loads, so a freshly opened
document has no source root yet; its `didSave` handler then panics on the main thread (notification
handlers are not panic-isolated, unlike requests) and the process exits. Confirmed from the 1.98
sources: `process_changes()` is gated on `vfs_done`, the `quiescent` flag of
`experimental/serverStatus` implies `vfs_done`, and a workspace-wide `cargo check` is kicked off on
becoming quiescent. rust-analyzer 1.97.1 panics the same way — CI had just been winning the race,
and the toolchain update shifted the timing. Closest upstream report:
rust-lang/rust-analyzer#19406 (same assertion, different trigger).

## Fix

- `🚑️ lsp:` opt into rust-analyzer's `experimental/serverStatus` notifications, track its
  `quiescent` flag, and hold `didSave` back while it is busy — it is sent on the document's next
  use instead, and rust-analyzer's own post-load `cargo check` covers the gap in the meantime.
- `🥅 lsp:` recover when rust-analyzer dies: in-flight requests fail as soon as its stdout closes
  instead of running into the 30 s timeout, the next tool call respawns it (liveness keyed on the
  reader task), a `start()` failing halfway cannot orphan the child (`kill_on_drop`), and a
  workspace diagnostics request on a dead rust-analyzer no longer falls back to an empty,
  clean-looking report. Verified by hand: `SIGKILL` rust-analyzer under a live server → the next
  call logs `rust-analyzer exited (signal: 9 (SIGKILL)), restarting it` and succeeds.
- `🔊 test-support:` / `🔊 lsp:` keep the MCP server's stderr in the daemon log (rotated past
  8 MiB) and relay rust-analyzer's stderr at `info`, so the next such death explains itself;
  `👷 ci:` uploads the daemon logs as an artifact of failed runs.
- Unit tests for all of it, driving the client against `cat`/duplex streams instead of a real
  rust-analyzer; TDD: the "no didSave while busy" test reproduced the old behaviour before the fix.

Verified locally with the CI invocation (`cargo test --locked --workspace --release
--all-features`) on rust-analyzer 1.98.0 — 61 tests pass across the workspace, no rust-analyzer
panics in the daemon logs — and the integration suite on 1.97.1 with fresh daemons; fmt and clippy
gates clean.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -59,3 +59,13 @@ jobs:
       - name: Build and Test
         run: |
           cargo test --locked --workspace --release --all-features
+      # The test daemons log to these files, rust-analyzer's own stderr included; they are the
+      # only trace of why a daemon's rust-analyzer died.
+      - name: Upload the test daemons' logs
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: test-daemon-logs
+          path: /tmp/rust-analyzer-mcp-sockets/*.log*
+          if-no-files-found: ignore
+          retention-days: 14
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -4,7 +4,7 @@ use serde_json::{json, Value};
 use std::{
     collections::{HashMap, HashSet},
     path::PathBuf,
-    process::Stdio,
+    process::{ExitStatus, Stdio},
     sync::Arc,
     time::Duration,
 };
@@ -12,6 +12,7 @@ use tokio::{
     io::{AsyncWriteExt, BufWriter},
     process::{Child, Command},
     sync::{oneshot, watch, Mutex},
+    task::JoinHandle,
 };
 
 use crate::{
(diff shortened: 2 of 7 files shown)


zeenix (commented):
🤖 ChatGPT GPT-5.6 on behalf of Zeeshan: Reviewed the failure analysis, lifecycle/restart handling, diagnostics behavior, tests, CI logging, and commit structure. The fix looks sound, the commits are appropriately split, and CI is green. I found no blocking issues; this is ready to merge.

PR #28, opened 2026-08-21
🐛 Never answer JSON-RPC notifications, and answer pings

### LINKED ISSUE #21: Server sends error response for `notifications/initialized`, causing transport closure
When an MCP client sends the `notifications/initialized` notification (part of the standard MCP initialization handshake), the server responds with:

```json
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found: notifications/initialized","data":null}}
```
This is incorrect behavior. `notifications/initialized` is a JSON-RPC **notification** (no `id` field). Per the JSON-RPC spec, notifications must not receive a response, including error responses. Unknown notifications should be silently ignored.

Any MCP client that strictly follows the spec will receive this unexpected error response and close the transport. The server then detects the broken pipe and shuts down, making it appear that the server exits immediately after initialization.

**Reproduction:**

```bash
printf '{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"capabilities":{},"clientInfo":{"name":"test","version":"1.0"},"protocolVersion":"2024-11-05"}}\n{"jsonrpc":"2.0","method":"notifications/initialized"}\n' \
  | rust-analyzer-mcp /path/to/workspace
```

The expected behavior is that the server silently ignores `notifications/initialized` (or handles it) and continues running, waiting for tool calls.

Fix: Before dispatching an incoming message to the method router, check whether it has an `id` field. If it does not, treat it as a notification and do not send any response, even on method-not-found.

zeenix:
Fixes #21.

`notifications/initialized` is part of every MCP handshake and, like any JSON-RPC notification, carries no `id`. `handle_request()` routed it through the method router anyway, so the server answered it with:

```json
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found: notifications/initialized","data":null}}
```

A client that follows the specification treats a response it never asked for as a protocol violation and closes the transport; the server then sees the broken pipe and shuts down, which is why it looked like it exited right after initialization.

## What changed

- **`src/mcp/server.rs`**: id-less messages are dropped before they reach the method router — unconditionally and method-agnostically, since an unknown *notification* gets no reply either way. Requests (which do carry an `id`) still get their `-32601`.
- **Second commit**: `ping` is answered with an empty result. It's a request every client may send at any time, including before `initialize`, to check the server is alive; it fell through to the catch-all and came back as "Method not found", which a client is free to read as a dead server. Say the word and I'll drop the commit if you'd rather keep this PR to the notification bug.

## Tests

`tests/integration/notifications.rs` drives the server binary directly — the shared test daemon pairs each request with one response, which is exactly what would hide a stray extra line. It feeds messages down stdin, closes it, and asserts on the exact set of lines that came back:

- `initialize` + `notifications/initialized` + `notifications/cancelled` + `tools/list` → exactly two responses, ids `0` and `1`.
- an unknown *request* still gets `-32601`, so the silencing cannot grow too broad.
- `ping` → `{}`.
- `{"id":null,...}` → nothing at all. MCP says the id "MUST NOT be null", so the message is malformed; answering it would mean sending back an id no client can match up. Today that falls out of `Option<Value>` mapping both spellings to `None` — the test pins the behaviour rather than leaving it incidental.

None of these start rust-analyzer, so they run in milliseconds. Verified the first one fails on `main` and passes here.

## Noticed but not touched

- `MCPResponse::Success` still has an optional `id` at the type level, so a future path could reintroduce an id-less response. Making it non-optional would rule that out by construction — a refactor, happy to do it separately.
- Tool failures come back as JSON-RPC errors with code `-1`; the tools spec would have execution failures come back as a result with `isError: true` so the model sees the message. Separate issue.

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -149,6 +149,15 @@ impl RustAnalyzerMCPServer {
                 continue;
             };
 
+            // A message without an `id` is a JSON-RPC notification, which must never be answered,
+            // not even with an error: a client that receives a response it did not ask for treats
+            // it as a protocol violation and closes the transport. `notifications/initialized` is
+            // part of every MCP handshake, so this used to break every spec-compliant client.
+            if request.id.is_none() {
+                debug!("Ignoring notification: {}", request.method);
+                continue;
+            }
+
             debug!("Received request: {}", request.method);
             // A shutdown signal must not wait for the request to finish: a tool call that
             // cold-starts rust-analyzer can run for minutes.
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -251,6 +251,13 @@ impl RustAnalyzerMCPServer {
                     }
                 }),
             },
+            // A liveness check the client may send at any point, including before `initialize`.
+            // Its result is empty; what matters is that it comes back at all.
+            "ping" => MCPResponse::Success {
+                jsonrpc: "2.0".to_string(),
+                id: request.id,
+                result: json!({}),
+            },
             "tools/list" => MCPResponse::Success {
                 jsonrpc: "2.0".to_string(),
                 id: request.id,
--- a/tests/integration/notifications.rs
+++ b/tests/integration/notifications.rs
@@ -44,6 +44,30 @@ async fn notifications_are_not_answered() -> Result<()> {
 }
 
 #[tokio::test]
+async fn ping_is_answered() -> Result<()> {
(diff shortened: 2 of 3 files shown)


zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed the JSON-RPC notification/request split, ping response, null-id behavior, tests, and green CI. No blocking findings.

PR #29, opened 2026-08-21
🐛 Convert between paths and file URIs properly

### LINKED ISSUE #16: hover, symbols, definition return null or error on Windows due to path handling bug
### Environment
- OS: Windows 11
- rust-analyzer-mcp version: v0.2.0
- rust-analyzer version: installed via `rustup component add rust-analyzer`

### Summary

On Windows, `rust_analyzer_hover`, `rust_analyzer_symbols`, and `rust_analyzer_definition` are non-functional. The root cause is that the server does not convert Windows paths to `file://` URIs for LSP requests, nor strip `file://` URIs back to plain paths for file I/O. Both conversions work trivially on Linux/Mac but require explicit handling on Windows.

`rust_analyzer_set_workspace` and `rust_analyzer_diagnostics` work correctly.

### Reproduction

Using the MCP Inspector (`npx @modelcontextprotocol/inspector rust-analyzer-mcp.exe`):

1. Call `rust_analyzer_set_workspace` with `C:\path\to\project` → succeeds
2. Call `rust_analyzer_hover` with the following path formats:

| Path format | Result |
|---|---|
| `C:\Users\...\file.rs` (Windows path) | Returns `null` silently |
| `\?\C:\Users\...\file.rs` (UNC path) | Server notification: `LSP error: {"code":-32603,"message":"url is not a file"}` |
| `file:///C:/Users/.../file.rs` (file URI) | `MCP error -1: Failed to read file: The filename, directory name, or volume label syntax is incorrect. (os error 123)` |

### Analysis

The server has two internal operations that both fail on Windows:

1. **LSP request** — expects a `file:///` URI with forward slashes. Receiving a raw Windows path causes a silent null or LSP error.
2. **File read** — expects a plain Windows path. Receiving a `file:///` URI causes OS error 123 because the URI is passed literally to the Windows filesystem API.

On Linux/Mac, `file:///home/user/foo.rs` ↔ `/home/user/foo.rs` is trivial. On Windows, `file:///C:/Users/.../foo.rs` ↔ `C:\Users\...\foo.rs` requires explicit conversion that is currently missing.

### Expected behaviour

`rust_analyzer_hover` (and other position-based tools) should accept a standard Windows path and work correctly, matching the behaviour of `rust_analyzer_diagnostics`.

zeenix:
Fixes #16.

A URI was built by pasting the path into `file://{path}` and a path was taken from a tool call as-is. Both work on Unix; neither does on Windows, where a path and a URI differ in every way that matters — the drive letter needs its own leading slash, the separators need turning around, anything outside the URI alphabet needs escaping. rust-analyzer was handed `file://C:\Users\...`, which it answered with `null` or rejected as `url is not a file`, so hover, symbols, definition, references and code actions were all dead there. Clients that worked around it by passing a `file:` URI as `file_path` then had that URI handed to the filesystem verbatim (`os error 123`).

## What changed

- **`src/uri.rs`** (new): `path_to_uri()`, `uri_to_path()` and `normalize()`, all on top of the `url` crate — the same crate rust-analyzer converts with, so the percent-encoding on both sides matches by construction. New dependency: `url = "2.5"`.
- **`src/mcp/server.rs`**: `resolve_path()` accepts a tool call's `file_path` spelled any of the three ways clients spell it — workspace-relative, absolute, or a `file:` URI — and `open_document_if_needed()` builds the document URI with `path_to_uri()`.
- **`src/lsp/client.rs`**: `rootUri` likewise.
- **`src/lsp/{connection,handlers}.rs`**: diagnostics are keyed by the URIs rust-analyzer publishes them under and looked up by the ones we build, so both sides now go through `normalize()`. That settles the one thing the `url` crate leaves open: on Windows the drive letter's case is whatever cargo happened to print, and `file:///c:/…` would otherwise never match `file:///C:/…`.
- **`src/mcp/handlers.rs`**: `rust_analyzer_set_workspace` takes a `file:` URI too.
- **README**: notes the three accepted spellings.

## Tests

- `src/uri.rs` unit tests: drive-letter normalisation, idempotence, the non-drive cases it must leave alone (UNC shares, `c:x`, non-file URIs), Unix round-trips through spaces, `#`, `?` and non-ASCII, plus `#[cfg(windows)]` round-trips including the `\\?\C:\…` form `canonicalize()` returns and `\\server\share`.
- `src/mcp/server.rs` unit tests for `resolve_path()` across the three spellings.
- `tests/integration/mcp_server_test.rs`: `rust_analyzer_symbols` called with a `file:` URI now returns symbols. Fails on `main` (the URI reaches `read_to_string()`), passes here.

## Second commit — Windows CI (drop it if you'd rather not)

This bug shipped because nothing in CI ever compiled, let alone ran, on Windows. `👷 ci: Build and unit-test on Windows too` adds a `windows-latest` job that builds and runs `cargo test --lib`. The integration tests can't follow — the shared test daemon talks over a Unix socket — which is why `test-support` moves under `[target.'cfg(unix)'.dev-dependencies]` (otherwise it gets built for the unit tests and fails to compile) and the four unit tests that stand a `cat`/`sh` process up in rust-analyzer's place become `#[cfg(unix)]`. I verified the Windows build of the lib and its unit tests with `cargo check --target x86_64-pc-windows-gnu --lib --profile test` under `-D warnings`; the `#[cfg(windows)]` URI tests are the ones the new job will actually run for the first time.

`cargo test --workspace` (18 unit + 18 integration tests) and `cargo clippy -- -D warnings` are green locally.

---

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -42,6 +42,25 @@ jobs:
         run: |
           cargo --locked clippy -- -D warnings
 
+  # The integration tests need a Unix socket and rust-analyzer, so Windows only gets the unit
+  # tests. Those cover the path and URI handling, which is what Windows gets wrong.
+  windows:
+    runs-on: windows-latest
+    needs: [fmt, clippy]
+    env:
+      RUSTFLAGS: -D warnings
+      RUST_BACKTRACE: full
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+      - uses: Swatinem/rust-cache@v2
+      - name: Build and run the unit tests
+        run: |
+          cargo --locked build
+          cargo --locked test --lib
+
   test:
     runs-on: ubuntu-latest
     needs: [fmt, clippy]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -446,6 +446,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1435fa1053d8b2fbbe9be7e97eca7f33d37b28409959813daefc1446a14247f1"
 
 [[package]]
+name = "dunce"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
+
+[[package]]
 name = "either"
 version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1706,6 +1712,7 @@ dependencies = [
  "arbitrary",
  "assert_cmd",
  "criterion",
+ "dunce",
  "env_logger 0.10.2",
  "futures",
  "jsonschema",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1723,6 +1723,7 @@ dependencies = [
  "test-support",
  "tokio",
  "tokio-test",
+ "url",
  "which",
  "wiremock",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ anyhow = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tokio = { version = "1.0", features = ["full"] }
+url = "2.5"
 env_logger = "0.10"
 log = "0.4"
 futures = "0.3"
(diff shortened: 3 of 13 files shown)


zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed path↔URI conversion on Unix and Windows, diagnostic-key normalization, dependency/CI changes, and the green Linux/Windows checks. No blocking findings.

Issue #12, opened 2026-01-11
Workspace diagnostics caching should be handled differently


ju6ge:
I am currently using this together with different models and testing it for code development purposes. During my tests I noticed the following behavior.

Model uses diagnostics to get errors -> Edits file to address issues -> Uses diagnostics again, gets the same response -> tries to read code to find issues but can not because the edit actually solved the issue, but the diagnostics are the same.

Looking at the code it seems that the the lsp client stores the results of the diagnostics:
https://github.com/zeenix/rust-analyzer-mcp/blob/ce8d64c9febe313bd9398ac9f6896a62e5097662/src/lsp/client.rs#L22-L31
https://github.com/zeenix/rust-analyzer-mcp/blob/ce8d64c9febe313bd9398ac9f6896a62e5097662/src/lsp/handlers.rs#L70-L100

But there it seems that updates to the source code are ignored completely thus makeing this problem a caching issue with this tools …

Expected behavior would be that getting diagnostics again should update the state. Now this is probably a bit more involved because as far as I am aware lsp is quite a chatty protocol when edits to files are made. So I am not entirely sure that a simple rust not cache and ask rust analyzer again would solve the issue.

But a fix is definitively needed for this otherwise there is no point in using this tool in an agentic development workflow. Because models will waist a lot of context until they figure out that the diagnostics are being cached!

PR #30, opened 2026-08-21
🐛 Tell rust-analyzer when an open document changed

zeenix:
Part one of #12. **Stacked on #29** (the base branch is #29's, so the diff shown here is only this change; it needs #29 merged first).

## The bug

`open_document()` sent `textDocument/didOpen` the first time a file was mentioned and never sent anything about it again. Once a document is open, rust-analyzer refuses to read it back off disk — that is the LSP contract, and its VFS enforces it explicitly ("if the file is in mem docs, it's managed by the client via notifications"). Its own file watcher sees the edit and drops it for exactly the files we opened.

So every answer about an edited file described the file as it looked when it was first mentioned: symbols without the function just written, a definition pointing at the line something used to be on, hover with the old type, diagnostics for code that no longer exists. With an agent at the other end — look, edit, look again — that is nearly every answer.

(There's a confusing asymmetry that made this hard to spot from the outside: `cargo check` reads the *disk*, so flycheck diagnostics could reflect the new content while everything served from rust-analyzer's own analysis reflected the old.)

## The fix

`open_documents` becomes a map of what rust-analyzer was last told about each document — version and a hash of the content. On each use:

- unseen → `didOpen` at version 1, as before;
- content hash differs → `didChange` at version + 1 carrying the whole file, then drop the diagnostics stored for that URI and re-arm the held-back `didSave` so the next check runs against the new content;
- unchanged → send nothing. This matters: rust-analyzer guards `didChange` with `if *data != new_contents`, so a redundant change never invalidates anything and never publishes anything. Sending one would mean waiting for a report that is not coming.

A whole-document change (one content-change event, no range) is what rust-analyzer's handler looks for first, so no diffing is needed even though it advertises incremental sync. Versions climb per document and reset with the process, which is what rust-analyzer echoes back in `publishDiagnostics.version` and checks in `codeAction/resolve`.

## Tests

- `src/lsp/client.rs`: four unit tests through the existing `cat`-as-rust-analyzer harness — a change is sent with the new text and version 2, an unchanged file sends nothing, versions climb 1/2/3, and an edit both drops the stored diagnostics and asks for another check.
- `tests/integration/edits.rs` (new): with its own isolated workspace, ask for symbols, append a function to the file on disk, ask again, and assert the new function is there. Fails on this PR's base (`["types", "utils", "run", …]` with no `added_after_opening`), passes here, and takes about a second.

`cargo test --workspace` and `cargo clippy -- -D warnings` are green.

## Next

Part two makes `rust_analyzer_diagnostics` wait for a cargo check that actually covers the new content, instead of the `file_path.contains("diagnostics_test")` heuristic it uses now. That needs the reader to stop mistaking rust-analyzer's own requests for responses first, so it comes as two more PRs on top of this one.

---

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -2,7 +2,8 @@ use anyhow::{anyhow, Result};
 use log::info;
 use serde_json::{json, Value};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{hash_map::DefaultHasher, HashMap, HashSet},
+    hash::{Hash, Hasher},
     path::PathBuf,
     process::{ExitStatus, Stdio},
     sync::Arc,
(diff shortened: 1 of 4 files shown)


zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed document version/hash tracking, didChange/didSave sequencing, diagnostic invalidation, and tests. The unit suite and edited-symbol integration test pass locally. No blocking findings.

PR #31, opened 2026-08-21
🐛 Tell rust-analyzer's requests apart from its answers

zeenix:
Groundwork for part two of #12. **Stacked on #30** (which is stacked on #29).

## Why

`handle_lsp_message()` treated everything that wasn't a notification as an answer to something we asked:

```rust
if json_value.get("method").is_some() && json_value.get("id").is_none() { … notification … }
let Ok(response) = serde_json::from_value::<LSPResponse>(json_value) else { return };
```

`LSPResponse` has no `deny_unknown_fields`, so a *request* from rust-analyzer (`{jsonrpc, id, method, params}`) parses as a response with `result: None` — and the code then does `pending.remove(&id)` and hands whoever is waiting on that id a `null`. The tool call behind it reports an empty result it never received.

rust-analyzer numbers its outgoing requests from 0 in the same space ours are numbered in (ours start at 1 and climb). It stays theoretical today only because rust-analyzer asks us exactly one thing — `workspace/configuration`, id 0. Part two of #12 declares `window.workDoneProgress`, which turns that into a `window/workDoneProgress/create` per progress run, and the collision stops being theoretical. Hence its own PR: it is a latent correctness bug either way.

## What changed

- **`src/lsp/connection.rs`**: messages are sorted by what they carry — method + id is a request, method alone a notification, id alone an answer — and requests are answered:
  - `workspace/configuration` → the settings the server was started with, once per section asked about. This is what makes the `workspace/didChangeConfiguration` we send at startup mean anything: rust-analyzer ignores what that notification carries and asks for the settings itself, so until now our configuration was dropped on the floor. (Answering with nothing would be worse than not answering: what comes back *replaces* the config rust-analyzer initialized with.)
  - `window/workDoneProgress/create` → `null`. Nothing to set up on our side; the answer is the point.
  - anything else → `-32601`, rather than being left unanswered, since rust-analyzer holds the request open otherwise. Never with a wrong or unsolicited id: `complete_request` panics its main loop on one of those.
- Answering means the reader task writes to rust-analyzer as well as reading from it, so the writing half is now an `Arc<Mutex<BufWriter<_>>>` shared with it, and the `Content-Length` framing both sides need moved next to it as `send_message()`.
- The settings themselves moved into one `settings()` used by `initializationOptions` and by the configuration answer, so the two cannot drift.

## Tests

Four new unit tests in `src/lsp/connection.rs`, driving the classifier over a duplex stream:

- a `window/workDoneProgress/create` with id 3 leaves a pending request with id 3 untouched and writes an answer with id 3 (this is the bug);
- `workspace/configuration` with two items is answered with two copies of the settings;
- an unsupported request is declined with `-32601` rather than dropped;
- an actual answer still reaches its waiter and clears the pending entry.

`cargo test --workspace` and `cargo clippy -- -D warnings` are green.

---

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -10,7 +10,7 @@ use std::{
     time::Duration,
 };
 use tokio::{
-    io::{AsyncWriteExt, BufWriter},
+    io::BufWriter,
     process::{Child, Command},
     sync::{oneshot, watch, Mutex},
     task::JoinHandle,
@@ -24,11 +24,13 @@ use crate::{
     uri,
 };
 
+use super::connection::{send_message, Outgoing};
+
 pub struct RustAnalyzerClient {
     pub(super) process: Option<Child>,
     pub(super) request_id: Arc<Mutex<u64>>,
     pub(super) workspace_root: PathBuf,
-    pub(super) stdin: Option<BufWriter<tokio::process::ChildStdin>>,
+    pub(super) stdin: Option<Outgoing<tokio::process::ChildStdin>>,
     pub(super) pending_requests: Arc<Mutex<HashMap<u64, oneshot::Sender<Value>>>>,
     pub(super) initialized: bool,
     /// What rust-analyzer was last told about each open document, keyed by normalized URI.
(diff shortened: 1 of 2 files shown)


zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed LSP request/notification/response classification, shared-writer serialization, configuration replies, and tests. The unit suite passes locally. No blocking findings.

PR #32, opened 2026-08-21
🐛 Report diagnostics from a check of the file as it is

### LINKED ISSUE #12: Workspace diagnostics caching should be handled differently

zeenix:
Closes #12 (with #30, merged, and #31). **Stacked on #31.**

## The bug

Whether `rust_analyzer_diagnostics` waited for anything was decided by:

```rust
let should_poll = file_path.contains("diagnostics_test") || file_path.contains("simple_error");
```

No file in this repository has matched that in a long time, so *every* call took the other branch: `sleep(2s)`, then return whatever happened to be in the published-diagnostics map. That is a report on the code as rust-analyzer last happened to see it — which, right after an edit, is the code as it was. #30 gets rust-analyzer the new content; this makes the tool wait for a report about it.

## What changed

**Wait for the workspace to load.** A file rust-analyzer hasn't reached yet looks exactly like a file with nothing wrong with it, so a report during loading can be an unqualified empty list. `fresh_diagnostics()` now waits for quiescence first (bounded).

**Cargo's half — wait for a check.** It asks for `rust-analyzer/runFlycheck` and waits for it. Waiting needs rust-analyzer to say what it's doing, so `initialize` declares `window.workDoneProgress` (without it `report_progress` returns early and sends nothing) and `connection.rs` follows the `$/progress` tokens under `rust-analyzer/flycheck/`. A check restarted while another runs *cancels* it, and a cancelled check ends exactly like a finished one — so the condition is "a check begun since the request, and none still running", not "an end arrived". There can be several checks (one per workspace), so the running ones are tracked as a set.

**rust-analyzer's own half — ask, don't wait.** This one caused a flaky failure I chased for a while: the analysis rust-analyzer publishes is whatever it last worked out, and an analysis begun before a change arrives *after* it, so the map could hold a description of the previous content even after a completed check. `textDocument/diagnostic` computes from the content rust-analyzer holds right now, so that half is pulled and the published entries are used only for what a check produced (`source != "rust-analyzer"`). If the request fails — older rust-analyzer, or one that has gone away — it falls back to the published entries.

**Bounded, and honest when it gives up.** A `runFlycheck` request is worked out on a task rust-analyzer abandons whenever its analysis is superseded (by the very change being asked about, among other things), so it can silently go nowhere; it is repeated a few times before giving up. Whenever any of the waits runs out, the result carries

```json
"note": "rust-analyzer had not finished loading the workspace or checking it when this was reported, so these diagnostics may be incomplete. Ask again for the rest."
```

rather than failing or passing what rust-analyzer had got to off as the state of the code.

Giving up on a rust-analyzer that reports no checks at all is not permanent (thanks for catching that): the skip applies only while no check has *ever* been reported, so the late first check on a big workspace puts the waiting back on the next call.

Removing the filename heuristic also removes the flat 2-second sleep every call used to pay.

## Tests

- `src/lsp/connection.rs`: the progress state machine — two concurrent checks, a cancelled-and-restarted check (which must *not* count as one that ran), and non-flycheck progress tokens being ignored.
- `src/lsp/client.rs`: a check is asked for and waited out; the analysis is asked for; a rust-analyzer that reports no checks comes back marked and isn't waited for again; **a check reported late puts the waiting back**; a report from a still-loading workspace is marked incomplete. All with `start_paused`, so no test waits on a real clock.
- `tests/integration/edits.rs::diagnostics_follow_an_edit`: with its own workspace, ask for diagnostics on a file full of deliberate errors, fix the file on disk, ask again, and assert the errors are gone.

I ran the whole suite repeatedly for this one — the earlier version passed alone and failed about half the time under the parallel load of the full run, which is what turned up the published-analysis staleness. `cargo test --workspace` (60 unit + 25 integration) and `cargo clippy -- -D warnings` are green.

## Note on `handle_request`'s dead paths

`RustAnalyzerClient::diagnostics()` still falls back to `textDocument/diagnostic` for `code_actions`, and `workspace_diagnostics()`'s `Err(_)` arm is unreachable (LSP errors are turned into `Ok(null)` before it can see them, and rust-analyzer has no `workspace/diagnostic` handler at all — the tool always reports "Unexpected response format"). Both are worth their own issue; I left them alone here.

### DIFF
--- a/src/lsp/handlers.rs
+++ b/src/lsp/handlers.rs
@@ -1,9 +1,50 @@
 use anyhow::Result;
-use log::info;
+use log::{info, warn};
 use serde_json::{json, Value};
+use std::time::Duration;
+use tokio::sync::watch;
 
-use super::client::RustAnalyzerClient;
-use crate::uri;
+use super::{client::RustAnalyzerClient, connection::Flycheck};
+use crate::{
+    config::{
+        DOCUMENT_OPEN_DELAY_MILLIS, FLYCHECK_REQUEST_ATTEMPTS, FLYCHECK_START_TIMEOUT_SECS,
+        FLYCHECK_TIMEOUT_SECS, WORKSPACE_LOAD_TIMEOUT_SECS,
+    },
+    uri,
+};
+
+/// What rust-analyzer marks the diagnostics it worked out itself with, as opposed to the ones it
+/// read out of a cargo check.
+const ANALYSIS_SOURCE: &str = "rust-analyzer";
+
+/// Diagnostics, and whether anything was still going on that could add to them.
+pub struct FreshDiagnostics {
+    pub items: Value,
+    /// False when the workspace was still loading, or the cargo check did not finish or never
+    /// started -- any of which makes these the best available rather than the last word.
+    pub complete: bool,
+}
+
+/// Waits for rust-analyzer's checks to reach `condition`, giving up after `timeout`.
+async fn wait_for(
+    progress: &mut watch::Receiver<Flycheck>,
+    timeout: Duration,
+    condition: impl Fn(&Flycheck) -> bool,
+) -> bool {
+    let wait = async {
+        loop {
+            if condition(&progress.borrow_and_update()) {
+                return true;
+            }
+            // Only fails once the client is gone, and then nothing is coming.
+            if progress.changed().await.is_err() {
+                return false;
+            }
+        }
+    };
+
+    tokio::time::timeout(timeout, wait).await.unwrap_or(false)
+}
 
 impl RustAnalyzerClient {
     pub async fn hover(&mut self, uri: &str, line: u32, character: u32) -> Result<Value> {
@@ -77,9 +118,9 @@ impl RustAnalyzerClient {
             "Available URIs with diagnostics: {:?}",
             diag_lock.keys().collect::<Vec<_>>()
         );
-        if let Some(diags) = diag_lock.get(&key) {
-            info!("Found {} stored diagnostics for {}", diags.len(), uri);
-            return Ok(json!(diags));
+        if let Some(diagnostics) = diag_lock.get(&key) {
+            info!("Found {} stored diagnostics for {}", diagnostics.len(), uri);
+            return Ok(json!(diagnostics));
         }
         drop(diag_lock);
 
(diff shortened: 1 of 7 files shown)


src/lsp/handlers.rs
```diff
@@ -101,6 +142,144 @@ impl RustAnalyzerClient {
...
+            self.send_notification("rust-analyzer/runFlycheck", Some(params.clone()))
+                .await?;
+        }
+        if !started {
+            info!("No cargo check started for {}, reporting what we have", uri);
+            // Having never seen one report a check, take it that this rust-analyzer does not,
+            // and stop making every later call wait the same wait out.
+            if before.never_ran_one() {
+                warn!("rust-analyzer does not report its cargo checks; not waiting for them");
+                self.reports_checks = false;
```
zeenix: 🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: This turns one missed start window into a permanent capability decision. On a large workspace the first requested check can begin after these five 2-second waits; the progress handler will then record it, but `reports_checks` stays false, so later diagnostics calls return immediately with incomplete cargo diagnostics. Please re-enable waiting once any flycheck begin has been observed, or keep this fallback scoped to the current call.

zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed flycheck state tracking, pull/publish diagnostic merging, bounded waits, and tests. The unit suite and edited-diagnostics integration test pass locally. One correctness issue remains below.

src/lsp/handlers.rs
zeenix: Right, and the amended commit makes it conditional rather than permanent: the skip is now `self.gave_up_on_checks && before.never_ran_one()`, so the moment the progress handler records any begin — including the late one from the check that call asked for — the next call waits again. `never_ran_one()` reads the live flycheck state, so nothing needs resetting.

New unit test `a_check_reported_late_puts_the_waiting_back` walks exactly that path: give up, then report a begin/end, then assert the following call waits its check out. It fails against the old sticky flag.

While in there I found the same "incomplete presented as complete" shape one step earlier and fixed it too: `fresh_diagnostics` now waits for the workspace to finish loading before asking for a check, because a file rust-analyzer hasn't reached yet looks exactly like a file with nothing wrong with it — an empty list with no qualification. The `from_a_finished_check` flag became `complete`, covering both the load and the check, and it's what the `note` in the result hangs off. That also fixed a flake where the first call in `diagnostics_follow_an_edit` reported zero errors for a file full of them.

---

zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Rechecked the amended commit. The earlier sticky flycheck fallback is fixed: once a late check is observed, the next diagnostics call waits again, and the new regression test exercises that exact path. The added workspace-loading timeout also marks potentially incomplete diagnostics honestly. No further findings.

PR #33, opened 2026-08-21
✨ Run rust-analyzer with the cargo features asked for

### LINKED ISSUE #13: How to pass --all-features to rust-analyzer?
Is there an flag for that?

zeenix:
Fixes #13. **Stacked on #32 → #31** — only for textual reasons (it rewrites the same `settings()`/`start()` lines those PRs touch); nothing in it depends on their behaviour.

## Why

Code behind a feature gate that isn't on is not part of the crate as far as rust-analyzer is concerned — no hover, no definitions, and no diagnostics about any of it. rust-analyzer takes its configuration over the LSP rather than from a command line, and this server sent it a fixed one, so there was no answer to the question in the issue.

## What it looks like

```
Options:
      --all-features         Analyse and check with every cargo feature enabled
      --features <LIST>      Cargo features to enable, comma- or space-separated. May be given
                             more than once
      --no-default-features  Leave the manifest's default features out
      --config <KEY=VALUE>   Set any rust-analyzer setting, such as --config check.command=clippy.
                             VALUE is read as JSON, or taken for a string if it is not JSON. May
                             be given more than once
```

The setup examples in the README stay default-neutral; the Cargo-features section is where the options are documented, and it says why the choice is per project — `--all-features` enables even features a crate never expects to see enabled together.

Details worth knowing:

- The features go into `cargo.features` / `cargo.noDefaultFeatures`, not `check.*`: the check settings fall back to the cargo ones, so this covers both what rust-analyzer analyses and what its `cargo check` checks, and it is also the spelling that works on rust-analyzer versions predating the `checkOnSave` → `check` rename.
- `--all-features` with `--features` or `--no-default-features` is refused rather than passed on — rust-analyzer would take all the features and silently ignore the rest.
- Features are split on commas *and* spaces and kept as separate list entries, because rust-analyzer splits each entry on `/` for the `package/feature` form; a joined `"a,b"` would break that.
- `--config` covers everything else (`check.command=clippy`, `cargo.target=…`, `check.extraArgs=["--tests"]`), so nobody has to wait for a flag. The value is parsed as JSON, or taken for a string when it isn't JSON.
- The composed settings are logged at startup: rust-analyzer says nothing at all about settings it doesn't recognise, so a typo would otherwise be undebuggable.

Argument parsing outgrew `main()` (it could only read one argument and had no way to express a flag), so it moved into `src/cli.rs` where it can be tested, with the settings it produces in `src/settings.rs`. `main.rs` is now just the dispatch.

## Tests

- `src/cli.rs`: 8 unit tests — `--features a,b` / `--features=a,b` / repeated / space-separated all reaching the same place, `--` ending the options, help and version winning, and every nonsensical command line being refused.
- `src/settings.rs`: 9 unit tests — the `"all"` literal being the lowercase bare string (a list would name a feature *called* `all`), features accumulating, the refused combinations, and `--config` creating paths through the settings and overriding what this server would otherwise say.
- `tests/integration/features.rs`: builds a throwaway crate whose only mistake is behind a feature gate, and checks the tool reports 0 errors by default and errors with `--all-features` and with `--features extra`. That is the whole issue in one assertion.

Verified by hand too, against the shipped rust-analyzer:

```
--- default features ---   {'errors': 0, 'hints': 1, 'warnings': 0}
--- all features ---       {'errors': 2, 'hints': 1, 'warnings': 0}
```

`cargo test --workspace` (60 unit + 25 integration) and `cargo clippy -- -D warnings` are green.

## One behaviour change worth flagging

The `checkOnSave` object the server used to send (`{enable, command, allTargets}`) is now just `checkOnSave: true`. In current rust-analyzer that setting is a boolean, and every value in the old object restated a default — it survived only through deprecated aliases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,45 @@ Add this to your Claude Desktop configuration (`claude_desktop_config.json`):
 **Note:** If installed from crates.io, the command will be in your PATH. For Claude Desktop, you
 may want to specify a `cwd` parameter if you want to analyze a specific project by default.
 
+### Cargo features
+
+By default rust-analyzer analyses and checks your crate with whatever features its manifest makes
+default, so code behind a feature gate looks unresolved and its errors go unreported. Pass the
+features to use on the command line:
+
+```bash
+rust-analyzer-mcp --all-features /path/to/project
+rust-analyzer-mcp --features serde,tokio /path/to/project
+rust-analyzer-mcp --no-default-features --features serde /path/to/project
+```
+
+In a workspace, a feature may need naming with the package it belongs to, exactly as cargo wants
+it: `--features my-crate/serde`.
+
+Which features to use is a choice worth making per project rather than everywhere: `--all-features`
+enables even the features a crate never expects to see enabled together, which for some projects
+means analysis that fails or takes far longer than it needs to. Options go in `args` alongside the
+command:
+
+```json
+{
+  "mcpServers": {
+    "rust-analyzer": {
+      "command": "rust-analyzer-mcp",
+      "args": ["--features", "serde,tokio"]
+    }
+  }
+}
+```
+
+Anything else rust-analyzer can be configured with is available through `--config`, which takes any
+[rust-analyzer setting](https://rust-analyzer.github.io/book/configuration.html) without the
+`rust-analyzer.` prefix. The value is read as JSON, or taken for a string if it is not JSON:
+
+```bash
+rust-analyzer-mcp --config check.command=clippy --config 'check.extraArgs=["--tests"]' .
+```
+
 ### Other MCP Clients
 
 For other MCP clients, run the server with:
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,0 +1,241 @@
+//! The command line this server is started with.
+
+use anyhow::{bail, Result};
+use std::{ffi::OsString, path::PathBuf};
+
+use crate::settings::Settings;
+
+pub const USAGE: &str = "\
+MCP server for rust-analyzer integration.
+
+Usage: rust-analyzer-mcp [OPTIONS] [--] [WORKSPACE]
+
+Arguments:
+  [WORKSPACE]  Path to the workspace root [default: current directory]
+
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -69,7 +69,8 @@ Add an MCP server configuration to one of these locations:
 {
   "mcpServers": {
     "rust-analyzer": {
-      "command": "rust-analyzer-mcp"
+      "command": "rust-analyzer-mcp",
+      "args": ["--all-features"]
```
zeenix: 🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: The generic setup now opts every Claude Code user into `--all-features`. That is not a safe default for projects with mutually exclusive features, and it can make analysis fail or become much more expensive. Please keep the base setup examples default-neutral and show `--all-features` only as an explicit choice in the Cargo-features section below.

zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed CLI parsing, settings composition, rust-analyzer configuration semantics, documentation, and focused feature tests. The unit and feature integration tests pass locally. One default-configuration issue is noted below.

README.md
zeenix: Agreed — the setup snippets are back to just `"command": "rust-analyzer-mcp"`, in all three places (both Claude Code options and Claude Desktop).

`--all-features` now appears only in the Cargo-features section, and that section says why it is a per-project choice: it enables even features a crate never expects to see enabled together, which for some projects means analysis that fails or takes far longer than it needs to. The `args` form is shown there instead, with `--features serde,tokio` as the example, so the mechanics are still documented without recommending the blunt option.

---

zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Rechecked the amended commit. The generic setup examples are default-neutral again; `--all-features` is now only an explicit per-project choice, with the mutually-exclusive-feature and cost caveats documented. No further findings.

PR #34, opened 2026-08-21
✨ mcp: Add a tool for renaming a symbol

### LINKED ISSUE #7: ✨/feat: support Rename Symbol
Supporting Rename Symbol would allow me to have an LLM refactor my projects and, for instance, convert tuple structs to named structs far more reliably.

zeenix:
Fixes #7. **Stacked on #33 → #32 → #31** (last of the chain).

Renaming is the refactoring an LLM is worst at and rust-analyzer is best at: it knows every use of a symbol, including the ones a text search would miss and the ones it would wrongly find.

## `🐛 lsp: Say what rust-analyzer refused to do` (first commit)

An LSP error was logged and then turned into a `null` result, so a request rust-analyzer *declined* was indistinguishable from one that found nothing. For rename that is the difference between "Invalid name `1`: not an identifier" and a rename that appears to have quietly done nothing. The message now reaches the caller as the tool call's error.

Two consequences worth naming:

- `workspace_diagnostics()`'s `Err(_)` fallback was unreachable before (errors never got there) and now runs — which is an improvement, since rust-analyzer has no `workspace/diagnostic` handler at all and that tool used to answer "Unexpected response format" every time.
- Every tool could now fail where it used to return `null` — including on `content modified`, which is rust-analyzer abandoning a request whose analysis was superseded mid-flight (any notification we send can do that). That is not a failure, it is "ask again", so `send_request` now retries it a few times. Tools are more robust than before this PR, not less.

## `✨ mcp: Add a tool for renaming a symbol` (second commit)

`rust_analyzer_rename` takes `file_path`, `line`, `character` and `new_name`, and **changes no files**: a server writing to a checkout takes the decision away from the host that asked and the person behind it, and the file may be open unsaved somewhere this server cannot see. What it does instead is everything needed to apply the edits correctly:

```json
{
  "applied": false,
  "old_name": "greet",
  "new_name": "welcome",
  "position_encoding": "utf-16",
  "summary": { "files_changed": 1, "edits": 3, "file_operations": 0 },
  "changes": [ { "file": "…/src/main.rs", "edits": [
      { "line": 45, "character": 19, "end_line": 45, "end_character": 24,
        "byte_range": [788, 793], "old_text": "greet", "new_text": "welcome" },
      … ] } ],
  "file_operations": [],
  "workspace_edit": { "documentChanges": [ … ] }
}
```

- **Byte ranges**, because the columns rust-analyzer speaks in are UTF-16 code units — neither bytes nor characters. They agree on ASCII lines and stop agreeing at the first `é` (byte offset runs ahead) or emoji (the code-unit count does).
- **`old_text`**, so an edit can be checked before it is made.
- **Edits last-first within each file**, so they can be applied one after another with no offset arithmetic. What rust-analyzer sends is ascending, which is exactly the order that is wrong to apply in.
- **`file_operations`**, because renaming a module renames its file. Asking for that requires declaring `workspace.workspaceEdit.resourceOperations`; without it rust-analyzer refuses the rename outright with "Client does not support rename capability".
- **`workspace_edit`** verbatim for anyone who would rather work it out themselves.

Two things a rename needs that a question about a single file doesn't:

1. **Every open document is brought up to date first.** A rename is worked out from what rust-analyzer holds for each file it touches; a stale one comes back as an edit to a line that has moved. A document whose file has since gone (what a module rename does to it) is closed, so it stops existing as far as rust-analyzer is concerned.
2. **The workspace has to have finished loading, and the tool fails if it hasn't.** A rename worked out halfway through misses the references rust-analyzer hasn't reached, and unlike a diagnostics report there is no way to hand back edits that are only safe to act on in part — whoever asked would have every reason to take them for the whole rename. `prepareRename` runs first too: its answer names the symbol being renamed (that's where `old_name` comes from), and it explains a position with nothing renameable better than the rename does.

## Tests

- `src/position.rs`: the UTF-16-column-to-byte-offset conversion, over `é` (1 unit, 2 bytes), `🦀` (2 units, 4 bytes), a column landing *inside* a character, positions past the end of a line and of the file, and CRLF endings.
- `src/lsp/connection.rs`: an error response reaching the waiter as the message rust-analyzer sent.
- `tests/integration/rename.rs`: a real rename returning `old_name`/`new_name`, every edit's `old_text` being the old name, the edits coming back last-first, and the file on disk being untouched; a rename to `1` failing with rust-analyzer's own words; a position with nothing to rename failing too.

The rename tests run against a server of their own (`test-project-rename`), because `test_workspace_change` points the shared one at a different workspace partway through the suite — which is what made them fail only in a full run until I chased it down.

`cargo test --workspace` (60 unit + 25 integration) and `cargo clippy -- -D warnings` are green.

## Deliberately not here

Applying the edits. If one-call refactoring is wanted, it should be a separate tool with its own name so the host can prompt for it — never folded into this one.

### DIFF
--- a/src/mcp/handlers.rs
+++ b/src/mcp/handlers.rs
@@ -1,10 +1,15 @@
 use anyhow::{anyhow, Result};
 use log::debug;
 use serde_json::{json, Value};
-use std::path::Path;
+use std::{
+    path::{Path, PathBuf},
+    time::Duration,
+};
 
 use crate::{
+    config::WORKSPACE_LOAD_TIMEOUT_SECS,
     diagnostics::format_diagnostics,
+    position,
     protocol::mcp::{ContentItem, ToolResult},
     uri,
 };
@@ -59,6 +64,7 @@ pub async fn handle_tool_call(
         "rust_analyzer_symbols" => handle_symbols(server, args).await,
         "rust_analyzer_format" => handle_format(server, args).await,
         "rust_analyzer_code_actions" => handle_code_actions(server, args).await,
+        "rust_analyzer_rename" => handle_rename(server, args).await,
         "rust_analyzer_set_workspace" => handle_set_workspace(server, args).await,
         "rust_analyzer_diagnostics" => handle_diagnostics(server, args).await,
         "rust_analyzer_workspace_diagnostics" => handle_workspace_diagnostics(server, args).await,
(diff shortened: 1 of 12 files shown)


zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Reviewed LSP error/retry plumbing, capability negotiation, open-document refresh, UTF-16-to-byte conversion, workspace-edit normalization, and tests. Formatting, clippy, and the unit suite pass locally. One blocking correctness issue remains below.

src/mcp/handlers.rs
```diff
@@ -212,6 +218,220 @@ async fn handle_code_actions(
...
+    };
+
+    // A rename is worked out from everything rust-analyzer has loaded, so one worked out while
+    // it is still loading can miss the references it has not reached yet -- and half a rename
+    // applied leaves the code worse than it was found.
+    if !client
+        .wait_until_loaded(Duration::from_secs(WORKSPACE_LOAD_TIMEOUT_SECS))
+        .await
+    {
+        warn!("rust-analyzer is still loading the workspace; renaming on what it has");
```
zeenix: 🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: The preceding comment correctly explains why this state is unsafe, but the timeout only logs and then returns a rename computed from a partially loaded workspace. A caller can apply that result believing it is workspace-wide and leave missed references broken. Please fail the tool call when `wait_until_loaded` returns false (or otherwise make it impossible to present this edit as complete).

zeenix: Fixed — the tool call now fails there:

```
rust-analyzer is still loading the workspace after 30s. A rename worked out now could miss
references, so it is not worth having; ask again once it has settled.
```

You are right that a note would not have been enough: unlike diagnostics, this result gets *applied*, and there is no way to hand back edits that are safe to act on only in part. The tool description and the README entry say the tool fails in that state, and the commit message no longer claims it renames on what rust-analyzer has.

The diagnostics tool has the same wait now (in #32, where `wait_until_loaded` moved down to), but keeps the note rather than failing: a report can honestly say "there may be more", and an empty list with no qualification was the actual bug there.

---

zeenix (commented):
🤖 ChatGPT GPT-5.6 Codex, on behalf of Zeenix: Rechecked the amended commits. A workspace-loading timeout now fails the tool call before `prepareRename`/`rename`, so a partial workspace edit cannot be presented as complete; the tool and README documentation match that behavior. No further findings.

PR #37, opened 2026-08-22
🔧 Show changelog emojis on section headings only

zeenix:
Every changelog entry repeated the emoji from its commit message, directly under a section heading that already said the same thing in words:

```
### Added
- ✨ Add a diagnostics tool backed by rust-analyzer.
```

This moves the emoji up to the heading, where it says something once:

```
### ✨ Added
- Add a diagnostics tool backed by rust-analyzer.
```

## What changed

Only `release-plz.toml`:

- **Group names carry the emoji** — `group = "Added"` became `group = "<!-- 00 -->✨ Added"`, rendered through `striptags | trim | upper_first`. The `<!-- NN -->` prefix pins section order, reproducing the order that previously fell out of the alphabetical sort of the bare group names, so no section moves.
- **Entries drop their emoji prefix** — one `replace_regex` both recognises and removes it, so the two cannot disagree about where the prefix ends. It strips **exactly one** gimoji plus the whitespace after it, matching the commit contract: a subject carries one prefix emoji, and everything after it — another emoji included — is title text. The alternation lists exactly the emoji this file routes to a group; any other emoji can only reach the catch-all, where nothing is stripped.
- **The prefix must be followed by whitespace.** A subject without a separator (`⚰️a Drop …`) stays visibly malformed rather than being mangled into `a Drop …`.
- **The catch-all group keeps its emoji.** `⚙️ Other` stands for no category, so an entry emoji there is information rather than repetition — 16 of this repo's 26 Other entries open with a distinct marker (`🔊`, `🔨`, `🚀`, `🚨`, `🤖`, `🧑‍💻`, `🧵`). Those entries render byte-identical to the pre-change output.
- **`upper_first` is guarded** — it applies only when nothing was stripped, and still sees the subject untrimmed. The emoji used to shield the rest of the subject from that filter; unguarded it rewrites identifiers like `zbus_macros shouldn't …` into `Zbus_macros shouldn't …`.
- **The dependency heading is 📦, not ⬆️.** That group also collects `➖`/`⬇️`/`📌`, which an upward arrow contradicts. Dependencies is the only group mixing opposite directions.

Commit preprocessors and the version-bump regexes are untouched. The emoji still sits at the start of the commit message, so parsers and bump rules match exactly as before — the strip happens at render time only.

## Verification

`git-cliff` 2.13.1 was run over the complete history of all eight repos sharing this template, with the pre-change and post-change configs, and every rendered entry compared as a multiset. In every repo: each entry in an emoji-headed group is its pre-change text minus exactly one prefix emoji, `Other` is byte-identical to pre-change, and the entry count is unchanged.

| repo | commits | entries |
|---|---|---|
| rust-analyzer-mcp | 132 | 111 |
| pixel8 | 502 | 408 |
| zbus | 6364 | 4987 |
| gimoji | 562 | 206 |
| busd | 637 | 296 |
| zlink | 1104 | 910 |
| zgvariant | 37 | 26 |
| zcheapstr | 21 | 10 |

Across all ~9,000 commits, exactly one subject carries two emoji (in zbus); per the contract its second emoji is title text and is kept. Nothing in this repo moves.

## Note

release-plz only prepends to `CHANGELOG.md`, so existing entries keep the old per-entry emoji. The new style starts with the next release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016iaTmaTUYJK5yuq86tGSuJ

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -36,16 +36,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 """
 
 # Changelog body template.
+# Each section heading carries the emoji of its category, so the leading emoji is stripped off
+# the entries rather than repeated on every line. One regex both recognises and removes it, so
+# the two can't disagree. It strips exactly one gimoji plus the whitespace after it, matching the
+# commit contract: a subject carries one prefix emoji, and everything after it — another emoji
+# included — is title text. So "✨ 🎉 Add foo" keeps its 🎉, and subjects with no separator after
+# the prefix ("⚰️a Drop …", "✨🏳️‍🌈 Rainbow") stay visibly malformed rather than being mangled.
+# The alternation is exactly the set of emoji this file routes to a group below; any other emoji
+# lands in the catch-all, where nothing is stripped. Keep the two in step when adding a parser.
+# The catch-all group is the exception: its heading emoji stands for no particular category, so
+# the entry emoji there is information rather than repetition and is kept. Those entries render
+# exactly as they did before this template existed.
+# `upper_first` applies only when nothing was stripped, and sees the subject untrimmed, exactly
+# as it did before. An emoji used to shield the rest of the subject from it; without that guard
+# it would rewrite "zbus_macros shouldn't …" into "Zbus_macros shouldn't …" and
+# "signature::Fields::len()" into "Signature::Fields::len()".
+# Group names are prefixed with an "<!-- NN -->" ordering comment, which `striptags` drops from
+# the rendered heading.
 # Issue refs are separated by "||" delimiter so we can insert period before them.
 body = """
 
 ## {{ version | trim_start_matches(pat="v") }} - {{ timestamp | date(format="%Y-%m-%d") }}
 {% for group, commits in commits | group_by(attribute="group") %}
-### {{ group | upper_first }}
+### {{ group | striptags | trim | upper_first }}
 {% for commit in commits %}\
 {% set first_line = commit.message | split(pat="\n") | first %}\
 {% set parts = first_line | split(pat="||") %}\
-- {{ parts | first | upper_first | trim | trim_end_matches(pat=".") }}\
+{% set subject = parts | first %}\
+{% if group is ending_with("Other") %}\
+{% set title = subject | upper_first %}\
+{% else %}\
+{% set title = subject | replace_regex(from="^\\s*(?:✨|🎉|🐛|🚑|🩹|🥅|⚡|📝|💡|♻|🎨|🚚|🔧|🏗|💄|💥|🔥|🗑|➕|➖|📌|⬆|⬇|🔒|✅|🧪|👷|💚)\\u{FE0F}?\\s+", to="") %}\
+{% if title == subject %}{% set title = title | upper_first %}{% endif %}\
+{% endif %}\
+- {{ title | trim | trim_end_matches(pat=".") }}\
 {% if commit.breaking %} (**BREAKING**){% endif %}.\
 {% if parts | length > 1 %} {{ parts | last }}{% endif %}
 {% endfor %}\
@@ -97,6 +121,8 @@ commit_preprocessors = [
 
 # Categorize commits based on gimoji emojis.
 # See: https://zeenix.github.io/gimoji/
+# The "<!-- NN -->" prefix on each group name fixes the order of the changelog sections
+# (the template strips it); the emoji right after it becomes the section heading emoji.
 commit_parsers = [
     # Skip commits carrying a "Changelog: skip" trailer (see the preprocessor above).
     { message = "^changelog-skip$", skip = true },
@@ -106,53 +132,53 @@ commit_parsers = [
     { message = "^[\\p{Emoji}\\p{Emoji_Presentation}\\p{Extended_Pictographic}]+$", skip = true },
 
     # Features.
-    { message = "^✨", group = "Added" },
-    { message = "^🎉", group = "Added" },
+    { message = "^✨", group = "<!-- 00 -->✨ Added" },
+    { message = "^🎉", group = "<!-- 00 -->✨ Added" },
 
     # Bug fixes.
-    { message = "^🐛", group = "Fixed" },
-    { message = "^🚑", group = "Fixed" },
-    { message = "^🩹", group = "Fixed" },
-    { message = "^🥅", group = "Fixed" },
+    { message = "^🐛", group = "<!-- 07 -->🐛 Fixed" },
+    { message = "^🚑", group = "<!-- 07 -->🐛 Fixed" },
+    { message = "^🩹", group = "<!-- 07 -->🐛 Fixed" },
+    { message = "^🥅", group = "<!-- 07 -->🐛 Fixed" },
 
     # Performance.
-    { message = "^⚡", group = "Performance" },
+    { message = "^⚡", group = "<!-- 09 -->⚡️ Performance" },
 
     # Documentation.
-    { message = "^📝", group = "Documentation" },
-    { message = "^💡", group = "Documentation" },
+    { message = "^📝", group = "<!-- 06 -->📝 Documentation" },
+    { message = "^💡", group = "<!-- 06 -->📝 Documentation" },
 
     # Refactoring/changes.
-    { message = "^♻️", group = "Changed" },
-    { message = "^🎨", group = "Changed" },
-    { message = "^🚚", group = "Changed" },
-    { message = "^🔧", group = "Changed" },
-    { message = "^🏗️", group = "Changed" },
-    { message = "^💄", group = "Changed" },
+    { message = "^♻️", group = "<!-- 03 -->♻️ Changed" },
+    { message = "^🎨", group = "<!-- 03 -->♻️ Changed" },
+    { message = "^🚚", group = "<!-- 03 -->♻️ Changed" },
+    { message = "^🔧", group = "<!-- 03 -->♻️ Changed" },
+    { message = "^🏗️", group = "<!-- 03 -->♻️ Changed" },
+    { message = "^💄", group = "<!-- 03 -->♻️ Changed" },
 
     # Breaking changes/removals.
-    { message = "^💥", group = "Breaking" },
-    { message = "^🔥", group = "Removed" },
-    { message = "^🗑️", group = "Deprecated" },
+    { message = "^💥", group = "<!-- 01 -->💥 Breaking" },
+    { message = "^🔥", group = "<!-- 10 -->🔥 Removed" },
+    { message = "^🗑️", group = "<!-- 05 -->🗑️ Deprecated" },
 
     # Dependencies.
-    { message = "^➕", group = "Dependencies" },
-    { message = "^➖", group = "Dependencies" },
-    { message = "^📌", group = "Dependencies" },
-    { message = "^⬆️", group = "Dependencies" },
-    { message = "^⬇️", group = "Dependencies" },
+    { message = "^➕", group = "<!-- 04 -->📦 Dependencies" },
+    { message = "^➖", group = "<!-- 04 -->📦 Dependencies" },
+    { message = "^📌", group = "<!-- 04 -->📦 Dependencies" },
+    { message = "^⬆️", group = "<!-- 04 -->📦 Dependencies" },
+    { message = "^⬇️", group = "<!-- 04 -->📦 Dependencies" },
 
     # Security.
-    { message = "^🔒", group = "Security" },
+    { message = "^🔒", group = "<!-- 11 -->🔒️ Security" },
 
     # Testing.
-    { message = "^✅", group = "Testing" },
-    { message = "^🧪", group = "Testing" },
+    { message = "^✅", group = "<!-- 12 -->✅ Testing" },
+    { message = "^🧪", group = "<!-- 12 -->✅ Testing" },
 
     # CI/Build.
-    { message = "^👷", group = "CI" },
-    { message = "^💚", group = "CI" },
+    { message = "^👷", group = "<!-- 02 -->👷 CI" },
+    { message = "^💚", group = "<!-- 02 -->👷 CI" },
 
     # Everything else.
-    { message = ".*", group = "Other" },
+    { message = ".*", group = "<!-- 08 -->⚙️ Other" },
 ]


release-plz.toml
```diff
@@ -36,16 +36,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## {{ version | trim_start_matches(pat="v") }} - {{ timestamp | date(format="%Y-%m-%d") }}
 {% for group, commits in commits | group_by(attribute="group") %}
-### {{ group | upper_first }}
+### {{ group | striptags | trim | upper_first }}
 {% for commit in commits %}\
 {% set first_line = commit.message | split(pat="\n") | first %}\
 {% set parts = first_line | split(pat="||") %}\
-- {{ parts | first | upper_first | trim | trim_end_matches(pat=".") }}\
+{% set title = parts | first | trim %}\
+{% if title is matching("^[^\\p{L}\\p{N}\\s]+\\s") %}\
```
zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: This predicate and the removal operation disagree. `\s` accepts tabs and other whitespace, while the next line splits only on literal spaces: `✨<TAB>Add after tab` renders as `After tab` (losing `Add`), and `✨<TAB>Tabbed` renders as an empty `- .`. It also strips load-bearing punctuation such as `# ` or `` ` ``, while missing `✨Add` and spaced multi-emoji prefixes. Please use one rendering-time, emoji-specific `replace_regex` operation for both recognition and removal; do not move this into a commit preprocessor, since parsers and bump rules still need the emoji.

release-plz.toml
```diff
@@ -36,16 +36,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+{% set title = title | split(pat=" ") | slice(start=1) | join(sep=" ") | trim %}\
+{% endif %}\
+- {{ title | upper_first | trim_end_matches(pat=".") }}\
```
zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: Moving `upper_first` past the stripped emoji changes case-sensitive names. A future `✨ rust-analyzer handles …` becomes `Rust-analyzer handles …`; previously `upper_first` saw the emoji and left the identifier untouched. Please preserve casing after a prefix was stripped, or apply `upper_first` only when the original title had no emoji prefix.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of zeenix: I independently rendered the old and new configs over the full history (110 entries before and after; no section moves). The ordering mechanism and Tera compatibility check out, but the two inline title-rendering issues are blocking. There is also a semantic loss in the new `Other` section: for example, `🔊 lsp: Log rust-analyzer's stderr at info level` becomes a plain `Log…` entry under `### ⚙️ Other`, even though `⚙️` does not replace the removed meaning. Please preserve the leading emoji for catch-all entries. Likewise, `⬆️ Dependencies` is misleading for the same group’s `➖` and `⬇️` entries; a neutral heading such as `📦 Dependencies` would fit.

release-plz.toml
zeenix: Agreed and fixed exactly as suggested — one rendering-time, emoji-specific `replace_regex`, kept in the template rather than a preprocessor for the reason you give (parsers and the bump regexes still need the prefix):

```
from="^\s*(?:[\p{Emoji_Presentation}\p{Extended_Pictographic}](?:\u{FE0F}|\u{200D})?\s*)*[\p{Emoji_Presentation}\p{Extended_Pictographic}](?:\u{FE0F}|\u{200D})?\s+"
```

Both of your synthetic cases reproduced before the fix — `✨<TAB>Tabbed` really did render as `- .`, because `split(pat=" ")` returned a single element and `slice(start=1)` emptied it. Whitespace is now consistently supported, `# ` and `` ` `` openers are no longer touched, and `✨ 🎉 …` is consumed whole.

One part I'm deliberately not taking: bare `✨Add` with no separator at all. Stripping without requiring whitespace turns malformed subjects like `⚰️a Drop …` (real, in a sibling repo) into `a Drop …`, and a stray letter is worse than a visible stray emoji. This repo has no such subjects, but the eight repos share this template and I'd rather it fail visibly than guess.

Verified over all 132 commits here: 111 entries, identical to the pre-change output apart from the removed emoji.

---

release-plz.toml
zeenix: Agreed — the emoji had been shielding the subject from `upper_first`, so moving the filter past the stripped prefix exposed every identifier. Fixed with your second option:

```
{% set title = subject | replace_regex(from=…, to="") %}
{% if title == subject %}{% set title = title | upper_first %}{% endif %}
```

`upper_first` now runs only on subjects that never had a prefix — exactly the set it used to affect — so a future `✨ rust-analyzer handles …` keeps its lowercase crate name.

This was not hypothetical in the sibling repos: `zbus_macros shouldn't …` became `Zbus_macros shouldn't …` and `signature::Fields::len()` became `Signature::Fields::len()`. Your comment also led me to a second instance of the same class — I had moved `trim` ahead of `upper_first`, which would have capitalised a subject with a leading space. The template keeps the original order now and lets the regex absorb leading whitespace instead.

Verified across all 111 entries here: zero casing differences from the pre-change output.

---

zeenix:
Both points from your review summary are taken — and I'd missed them until now, having only read the inline threads and not the review bodies. Apologies for the extra round.

**Emoji preserved in the catch-all group.** You're right that `⚙️` replaces no meaning, and it was broader than the one example: **16 of the 26 Other entries** here open with a distinct emoji (`🔊`, `🔨`, `🚀`, `🚨`, `🤖`, `🧑‍💻`, `🧵`). `🔊 lsp: Log rust-analyzer's stderr at info level` reproduces exactly as you describe.

The rule is now: strip the emoji where the heading already carries it, keep it where the heading stands for no category. That is precisely the "repetition, not information" distinction this change is about, and Other was the one group where I had it wrong. Those entries now render **byte-identical to the pre-change output** — including `upper_first` behaviour, since with the emoji retained the filter is a no-op there exactly as it always was.

**Neutral dependency heading.** Agreed; now `📦 Dependencies`. This repo has only one Dependencies entry today, but the group accepts `➖`/`⬇️`/`📌` and the template is shared across eight repos where 38 such entries exist, so the upward arrow was misstating them. One overlap worth flagging: `📦️` is itself a gimoji (compiled files/packages) and shows up as an entry marker in the Other group of sibling repos. In heading position I think it reads as "packages", but say the word and I'll pick something else.

I checked whether any other group has the same defect — mixing opposite directions under one heading. Dependencies is the only one; Fixed (`🐛🚑🩹🥅`), Changed (`♻️🎨🚚🔧🏗️💄`), Added (`✨🎉`), CI (`👷💚`) and Testing (`✅🧪`) each point one way, so their representative emoji are accurate.

Re-verified over all 132 commits after these changes: 111 entries, Other byte-identical to before, every other section identical apart from the removed emoji.

---

release-plz.toml
```diff
@@ -36,16 +36,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### {{ group | striptags | trim | upper_first }}
 {% for commit in commits %}\
 {% set first_line = commit.message | split(pat="\n") | first %}\
 {% set parts = first_line | split(pat="||") %}\
-- {{ parts | first | upper_first | trim | trim_end_matches(pat=".") }}\
+{% set subject = parts | first %}\
+{% if group is ending_with("Other") %}\
+{% set title = subject | upper_first %}\
+{% else %}\
+{% set title = subject | replace_regex(from="^\\s*(?:[\\p{Emoji_Presentation}\\p{Extended_Pictographic}](?:\\u{FE0F}|\\u{200D})?\\s*)*[\\p{Emoji_Presentation}\\p{Extended_Pictographic}](?:\\u{FE0F}|\\u{200D})?\\s+", to="") %}\
```
zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: The previous split/whitespace issue is fixed, but this replacement is still lossy. `Extended_Pictographic` includes text-default, load-bearing symbols: exact git-cliff 2.13.1 renders `✨ © 2026 licenses` as `- 2026 licenses.`, `✨ ™ Product` as `- Product.`, and `✨ ↩ Return behavior` as `- Return behavior.`. The atom also allows FE0F *or* ZWJ, not the common FE0F+ZWJ sequence: `✨ ❤️‍🔥 Heart` leaves `❤️‍🔥`, while attached `✨🏳️‍🌈 Rainbow` strips nothing; keycaps/tag flags have analogous gaps. Please use a repeated explicit alternation of the configured gimoji sequences (safest). Alternatively, model a complete emoji-presentation sequence and require VS16 before treating text-default pictographs as emoji.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of zeenix: Rechecked the amended head with the exact renderer and full history. The casing guard, catch-all preservation, `📦 Dependencies`, ordering, and repository-specific settings are correct; no entries disappeared or moved. One new blocking edge in the replacement regex remains below.

release-plz.toml
zeenix: Confirmed, on the same git-cliff 2.13.1. All three reproduce, and two more that you didn't list are worse:

| subject | before | after |
|---|---|---|
| `✨ © 2026 licenses` | `- 2026 licenses.` | `- © 2026 licenses.` |
| `✨ ™ Product` | `- Product.` | `- ™ Product.` |
| `✨ ↩ Return behavior` | `- Return behavior.` | `- ↩ Return behavior.` |
| `✨ ❤️‍🔥 Heart` | `- ❤️‍🔥 Heart.` | `- Heart.` |
| `✨🏳️‍🌈 Rainbow` | `- ✨🏳️‍🌈 Rainbow.` | `- Rainbow.` |
| `🐛 ™ Trademark mid-title` | `- Trademark mid-title.` | `- ™ Trademark mid-title.` |

The attached-rainbow case left even the leading `✨` in place, and `🐛 ™ …` destroyed a trademark sign in an ordinary Fixed entry — not a decorative prefix at all.

I took your second option, modelling a complete emoji-presentation sequence:

```
^\s*(?:(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F})(?:\u{200D}(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F}))*\s*)*(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F})(?:\u{200D}(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F}))*\s+
```

A text-default pictograph only counts as emoji when it carries U+FE0F, which is what rules out `©`, `™` and `↩`; ZWJ now joins whole sequences rather than being an alternative to FE0F; and the optional FE0F after a default-emoji codepoint keeps `⚡️` working where the parser matches bare `^⚡`.

Three reasons I went with this over the explicit alternation you called safest — happy to switch if you disagree:

1. **It fixes a case the alternation would not.** `✨🏳️‍🌈 Rainbow` strips cleanly here; against a list of the configured gimoji, `🏳️‍🌈` isn't in the list, so the run would break at `✨` and nothing would be stripped — exactly the wart you flagged.
2. **No list to keep in sync.** The eight repos share this template but have slightly different parser sets, so an explicit alternation means eight lists that must track eight `commit_parsers` blocks.
3. **The failure mode is conservative.** Anything the pattern doesn't recognise is left in place. Keycaps and tag sequences still aren't stripped — `✨ 1️⃣ Keycap` renders as `1️⃣ Keycap` — so the worst case is a visible unstripped emoji, never destroyed text. That asymmetry is what went wrong before.

Verified over all 132 commits here, and over all 6,954 entries across the eight repos: output identical to the previously reviewed render. The tightening only changes the synthetic edge cases; no real entry moved.

One pre-existing note: the same over-broad atom lives in this repo's scope-stripping `commit_preprocessor`, which I copied it from. It's lower risk there (it also requires a `word: ` scope to follow), but I'm happy to tighten it in a follow-up if you want them consistent.

---

zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: Clarification from the maintainer: the project contract is **exactly one prefix emoji, drawn from the curated gimoji set**. Anything after that prefix—including another emoji—is ordinary title text. My earlier request to model arbitrary compound emoji runs was therefore overbroad.

The current amended regex is still wrong under that contract because its repeated group consumes more than one emoji sequence. For example:

- `✨ 🎉 Add foo` must render as `- 🎉 Add foo.`, not `- Add foo.`
- `✨ ❤️‍🔥 Heart behavior` must render as `- ❤️‍🔥 Heart behavior.`, not `- Heart behavior.`
- `✨🏳️‍🌈 Rainbow` has no separator after the one allowed prefix and should remain visibly malformed rather than being consumed as a combined run.

Please strip **exactly one** recognized gimoji plus its required following whitespace. An explicit alternation of the curated prefixes is appropriate here; there is no need to recognize or repeat over arbitrary Unicode emoji sequences.

zeenix: Understood, and done — the contract settles it. It also retires my main argument for the presentation model: I justified it partly because it stripped `✨🏳️‍🌈 Rainbow` where an alternation would not, but under "one prefix, rest is title text" that subject *should* stay unstripped. I was optimising for the wrong target.

Now exactly one gimoji plus its required whitespace:

```
^\s*(?:✨|🎉|🐛|🚑|🩹|🥅|⚡|📝|💡|♻|🎨|🚚|🔧|🏗|💄|💥|🔥|🗑|➕|➖|📌|⬆|⬇|🔒|✅|🧪|👷|💚)\u{FE0F}?\s+
```

Your three cases, rendered on git-cliff 2.13.1:

| subject | renders as |
|---|---|
| `✨ 🎉 Add foo` | `- 🎉 Add foo.` |
| `✨ ❤️‍🔥 Heart behavior` | `- ❤️‍🔥 Heart behavior.` |
| `✨🏳️‍🌈 Rainbow` | `- ✨🏳️‍🌈 Rainbow.` |

The alternation is the exact set of emoji this file routes to a group, with the variation selector optional so bare-parser forms like `^⚡` still strip `⚡️`. It doesn't need to be wider: an emoji that isn't in that list can only reach the catch-all, and nothing is stripped there. I've noted in the config comment to keep the two in step when adding a parser — and if they ever drift, the failure is a visibly unstripped emoji, not lost text.

Across all ~9,000 commits in the eight repos, exactly one real entry changes as a result: zbus's `🩹 🚸 zv: \`Signature\` equality regardless of outer parentheses` now keeps its `🚸` as title text. Nothing in this repo moves.

Verified across all eight: every entry in an emoji-headed group is now its pre-change text minus exactly one prefix emoji, Other is byte-identical to pre-change, and entry counts are unchanged (111 here).

---

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of zeenix: Rechecked the latest amended head against the strict one-curated-gimoji-prefix contract. Exact git-cliff rendering over the full history preserves all 111 entries and their text apart from the intended prefix removal/heading change; extra or unknown emoji after the prefix remain title text, and CI is green. No remaining issues—ready to merge.

PR #39, opened 2026-09-12
🐛 test-support: fix two macOS-only IPC socket bugs

IconRunner05:
Two macOS-only bugs in the IPC test harness, both pre-existing and both found while adding an unrelated test. Each is a few lines.

**1. An accepted connection inherits O_NONBLOCK from the listener.**

The listener is deliberately non-blocking so the accept loop can poll for shutdown. On the BSDs — macOS among them — `accept()` hands the new connection that flag too; on Linux it does not, which is why CI never saw this. Everything that serves a client then treats a non-blocking socket as blocking, so a response larger than the socket's send buffer stops after 8192 bytes and fails with `EAGAIN`. The client parses what arrived:

```
Error: EOF while parsing a string at line 1 column 8192
```

and the daemon logs `Client connection error: Resource temporarily unavailable (os error 35)` and carries on. Whether a test passed depended only on how big the answer was, so it stayed hidden until a tool answered with more than 8 KiB.

**2. The daemon's socket path does not fit in `sun_path`.**

`sun_path` is 104 bytes on macOS including the terminator, and a `TMPDIR` there is already about half of that (`/var/folders/xx/<26 characters>/T/`). With the directory spelled `rust-analyzer-mcp-sockets`, the socket for `test-project-diagnostics` came to exactly 104 bytes, so its daemon could never bind — the four diagnostics tests each failed after a full minute with `Failed to connect to MCP server after starting`, and the reason (`path must be shorter than SUN_LEN`) only ever reached the daemon's own log file. Renaming the directory `ra-mcp-socks` leaves 13 bytes of headroom.

Measured on macOS 15 (arm64), `cargo test`: before, 4 diagnostics tests failed to connect and both `shared_test` tests failed on the truncation; after, the whole suite passes.


test-support/src/ipc/server.rs
```diff
@@ -376,8 +384,15 @@ fn wait_for_ready(
...
+/// Where the daemon for `project_type` listens.
+///
+/// The directory is named as briefly as it can stand being: a unix socket's path has to fit in
+/// `sun_path`, which is 104 bytes on macOS including the terminator, and a temporary directory
+/// there is already about half of that (`/var/folders/xx/<26 characters>/T/`). Spelling the
+/// directory out in full put the longest project's socket at exactly 104 bytes, and every test
+/// using it failed to reach a daemon that could never bind.
 pub fn socket_path(project_type: &str) -> PathBuf {
-    let socket_dir = std::env::temp_dir().join("rust-analyzer-mcp-sockets");
+    let socket_dir = std::env::temp_dir().join("ra-mcp-socks");
```
zeenix, line 395: Reviewed by Codex on behalf of zeenix.

[P3] Update the CI log artifact path with the socket directory

The failure-log upload in `.github/workflows/rust.yml` still reads
`/tmp/rust-analyzer-mcp-sockets/*.log*`. Daemon logs now live beside the sockets under
`/tmp/ra-mcp-socks`, so CI failures will silently upload no logs (`if-no-files-found: ignore`).
Please update that path in the socket-directory commit.

zeenix (commented):

The two socket fixes look sound by inspection, and they are correctly split into atomic
commits. This is a foundation for #41 and the later stack. One small CI follow-up is inline;
CI itself is still awaiting workflow approval.

PR #40, opened 2026-09-12
🐛 mcp: wait for the index before answering about a symbol

IconRunner05:
`wait_until_loaded` is already here, already fed by `experimental/serverStatus`, and already used by the two paths that would be actively harmful without it — the diagnostics path, and `handle_rename`, which refuses rather than hand back half a rename. The read tools use none of it.

So `hover`, `definition`, `references`, `completion` and `code_actions` answer from whatever rust-analyzer has loaded at the time, which for the first stretch of a large workspace is `null` and `[]`. Measured on a ~2 GB workspace: 45 seconds after start, every index-backed call was empty at a position later *proven* correct; warm, the same call returned the full signature and 31 references. Nothing in the response distinguishes the two, and `null`/`[]` is also what these tools answer for a symbol nothing refers to and for a position that is not on a symbol — so "the index is not ready yet" is indistinguishable from "this code is unused", which is the one of the three anybody acts on.

This waits on the same gate the other paths use, and when the wait runs out it says so instead of answering emptily:

> rust-analyzer is still loading the workspace after 30s. An answer worked out now would be from a partial index, and an empty one could not be told from a symbol that is really unused; ask again once it has settled.

`rust_analyzer_symbols` is left ungated deliberately, with a comment saying why: the symbols of one file are worked out from that file alone, so it answers correctly while the workspace is still loading — and it is the one useful thing left to ask while everything else is waiting.

No new LSP method, no new dependency; the mechanism was already built and proven.


zeenix (commented):

No implementation blocker found in this change. The five index-dependent read paths use the
existing loading gate, propagate its timeout, and leave document symbols ungated as intended.
The combined stack builds, passes all 60 library tests, and passes the live navigation checks.
This can land independently once CI runs.

PR #41, opened 2026-09-12
🐛 lsp: ask for the symbol shape that says where the name is

IconRunner05:
`initialize` declares `"documentSymbol": {}`, which says nothing about understanding the nested `DocumentSymbol` shape — so rust-analyzer answers `textDocument/documentSymbol` in the legacy flat `SymbolInformation` form. That form reports each item's **whole** range, doc comment and attributes included, and carries no `selectionRange` at all.

The consequence is that a symbol's reported position is not its name. Measured against this repo's own `test-project/src/lib.rs`:

| symbol | flat `range.start.line` | actual identifier line |
| --- | --- | --- |
| `run` | 6 (its `///` line) | 7 |
| `Person` | 14 (its `#[allow(dead_code)]`) | 15 |
| `add` | 35 (the comment above it) | 36 |

That matters because the only thing a consumer can do with a symbol's position is ask another question at it — and `hover`, `definition` and `references` at a doc comment answer `null` and `[]`, which is also exactly what they answer for a symbol nothing uses. So a client reading the flat shape gets wrong positions and has no way to tell that it has.

Declaring `hierarchicalDocumentSymbolSupport` gets the nested shape, whose `selectionRange` is the identifier itself — line **and** column, so a caller no longer has to count columns by hand — which nests members under the item they belong to, and which carries each signature in `detail`.

Adds `tests/integration/symbols.rs`: every symbol must carry a `selectionRange` whose line actually contains its name, and `run` specifically must report its `range` at the doc comment and its `selectionRange` at `fn run`, with the character landing on the `r`.

Builds on #39 — the nested payload for that one file is over 8 KiB, which is precisely what the truncation bug there swallows, so the new test cannot pass without it. Review the last commit only if that one lands first.


zeenix (commented):

No blocker found in the distinct document-symbol change. A live protocol check confirms the
nested shape and `run`'s identifier in `selectionRange` on line 7, separate from its doc comment.
Depends on #39. Please carry forward the null-result retry currently introduced in #45 before
considering the document-symbol cold-start behaviour fully covered.

PR #42, opened 2026-09-12
🐛 mcp: stop a workspace nobody loaded from reading as dead code

IconRunner05:
Two halves of one bug: rust-analyzer knows exactly one workspace, and about a file outside it, it does not say "I have not loaded that" — it says `null`, or `[]`. That is the same answer it gives for a symbol nothing refers to and for a position that is not on a symbol, so pointing the server at the wrong place produces confident, quiet nonsense: every file in the project reads as dead code.

**`set_workspace` accepted anything.** A directory with no `Cargo.toml` came back as `Workspace set to: …` and every later question was answered from an index of nothing. It also shut the existing client down *before* looking at the new path, so a typo cost the session its working server. Now it checks for a manifest first — taking `Cargo.toml` itself to mean the directory holding it — and refuses before tearing anything down, naming the workspace it is leaving in place.

**An empty answer about a file outside the loaded workspace now says so**, and names the workspace that *is* loaded. Only when the answer was empty anyway: a file outside the root is perfectly askable when rust-analyzer has it loaded (following a definition into a dependency's sources, say), and that still answers exactly as before.

Reproduced before the fix: a client whose working directory had no Rust in it got a bare `null` from `hover` at a position that had returned a full signature minutes earlier, because a respawn had reset the workspace to that directory.

Two test-harness bugs found on the way, in their own commits:

- **The tests can run against a stale binary.** Both places that start the server preferred `target/release/rust-analyzer-mcp` whenever it existed, and `cargo test` never rebuilds it. Two new tests in this PR failed against a release binary built one branch earlier, reporting the old behaviour in full. Now the profile the tests were built with is preferred, with a warning on stderr when only the other exists.
- **`test_workspace_change` retargeted the shared daemon** and left it retargeted, so later tests on that daemon asked about `test-project` files while rust-analyzer was loaded somewhere else — and got `null`, which those tests accept as valid null-handling. It gets a daemon of its own. This surfaced *because* of the fix above: the new error turned a vacuous pass into a loud failure.


src/mcp/handlers.rs
```diff
@@ -445,16 +526,36 @@ async fn handle_set_workspace(
...
+    let workspace_root = match named.file_name() {
+        Some(name) if name == "Cargo.toml" => named.parent().unwrap_or(&named).to_path_buf(),
+        _ => named,
+    };
+
+    // A directory with no manifest in it is not a workspace, and rust-analyzer started on one
+    // has nothing loaded and says nothing about any file -- which reads as a workspace full of
+    // code nothing refers to. Refusing here costs a typo'd path; accepting one quietly redirects
+    // every question asked afterwards.
+    if !workspace_root.join("Cargo.toml").is_file() {
```
zeenix, line 542: Reviewed by Codex on behalf of zeenix.

[P2] Preserve rust-analyzer's supported workspace discovery

A `Cargo.toml` directly inside the requested directory is not required for rust-analyzer to
load a project. Reproduced with a normal Cargo project: starting this server in its `src/`
directory returns the correct hover for `lib.rs`, but `set_workspace` to that same working
directory is now refused by this check. rust-analyzer discovers parent manifests, child
projects and `rust-project.json`; the existing `--config` also permits `linkedProjects`.

Please validate using the effective project/discovery configuration rather than requiring this
one file at the root. Otherwise this removes previously working workspaces, including
[non-Cargo projects](https://rust-analyzer.github.io/book/non_cargo_based_projects.html).

zeenix (changes requested):

The harness fixes are separately committed as requested. The new workspace validation rejects
working rust-analyzer configurations, however; details and a reproduced Cargo case are inline.
Actual ancestry includes #39, #40 and #41. Fix this here and propagate it through #45 onward.

PR #43, opened 2026-09-12
✨ mcp: add a tool to find a symbol by name

IconRunner05:
Every tool in this server takes a file, a line and a character — and nothing in it produces one. A caller who knows a symbol's name has to find its position by searching the text of the files first, which cannot tell a definition from the same word in a comment, a string, or a `#[cfg(test)]` fixture, and which is the step that silently makes every later answer wrong when it lands a few lines off.

`workspace/symbol` is the LSP method for exactly this, and rust-analyzer has advertised `workspaceSymbolProvider` all along. Each result's `location.range` is the identifier itself, so it drops straight into the position the other tools want — verified in the test added here, which checks the reported position actually lands on the name.

The tool's description says at length what the method is and is not: the query is matched fuzzily, results are scored, only the best few come back, and rust-analyzer searches its index rather than the files. It finds a symbol you can name; it never proves one absent and it counts nothing. That belongs in the description rather than the docs, because the caller is usually a model reading the description, and read as a census this is wrong in the direction that looks right.

Requires `workspace.symbol` in the client capabilities — without it rust-analyzer answers nothing at all.


zeenix (commented):

No blocker found in the distinct workspace-symbol implementation. Live testing found
`Calculator` and returned its identifier location; request wiring, capability and readiness
handling are consistent. Actual dependencies are #39/#41 plus #40. Land those first, then
rebase this to expose only its own change.

PR #44, opened 2026-09-12
✨ mcp: answer the questions text search cannot

IconRunner05:
Four LSP methods rust-analyzer has advertised all along and this server never asked for. Each is a capability, a client method, a handler and a tool.

- **`textDocument/implementation`** — what implements the trait, or the trait method, at a position. The one question no text tool can answer even in principle: which type carries an implementation is decided by dispatch, not by anything written where the question is asked, so grep and structural search miss it entirely rather than partially.
- **`textDocument/typeDefinition`** — where the *type* of the thing at a position is defined, as opposed to the thing itself.
- **`callHierarchy/incomingCalls`** and **`outgoingCalls`** — which *functions* call this one, and which it calls, as distinct from where its name appears. `references` answers the second question and is routinely read as answering the first.

The two call tools resolve the call-hierarchy item themselves rather than making the caller run `prepareCallHierarchy` and hand the item back — the item is of no use for anything else. It is returned alongside the calls so what was actually asked about is visible, and a position on nothing callable is an error saying so rather than an empty list.

Tests cover all four against `test-project`: `run` appears among the callers of `process`, `validate` among its callees, a doc comment is refused as a call-hierarchy position, and `Default` resolves to its implementation.

Builds on #42 (the shared-daemon contamination fix): without it, `test_workspace_change` leaves the shared daemon pointed at a temporary copy and the new call-hierarchy tests fail on whichever order the suite happens to run in. Stacked on #43 for the same workspace-capability block in `initialize`; review the last commit only.


zeenix (commented):

The new implementation, type-definition, incoming-call and outgoing-call paths passed live
protocol checks; no additional implementation defect found. Please split `e758222` into atomic
commits for implementation lookup, type-definition lookup, and call hierarchy: those are
independently usable changes, despite sharing the same plumbing.

This branch contains #43's ancestry but does not actually contain #42. Its stated test-harness
dependency on #42 therefore still needs to land first.

zeenix:
@IconRunner05 please rebase. Also I'm execting you to get all your PRs to be merged.

PR #45, opened 2026-09-12
✨ mcp: expose the rust-analyzer extensions worth having

IconRunner05:
Three requests that are rust-analyzer's own rather than LSP, and that nothing else substitutes for.

- **`rust-analyzer/expandMacro`** — what a macro call becomes. Macro-generated code is in no file, so reading the source cannot show it however it is read, and hover does not show it either. The only way to see where a type, method or impl that appears from nowhere comes from.
- **`rust-analyzer/relatedTests`** — which tests reach a symbol. Worth knowing before changing one.
- **`experimental/runnables`** — the exact cargo invocation for a file, or for the single test at a position, with the arguments and environment rust-analyzer would use.

All three were probed against rust-analyzer 1.97.1 before being wired, which was worth doing twice over:

- `expandMacro` and `relatedTests` are **not** listed in the `experimental` block of the `initialize` response, and are answered anyway. The advertised capability list understates what is there, so absence from it is not evidence of absence.
- Runnables are advertised as `runnables`, and `rust-analyzer/runnables` is `unknown request`: the method is `experimental/runnables`. The two namespaces are not interchangeable, and nothing reported by the server says which extension lives in which.
- `textDocument/prepareTypeHierarchy` really is absent (`unknown request`), so type hierarchy is not wired here.

Deliberately left out: `experimental/ssr` (answers, but produces workspace edits — an edit-producing tool wants the treatment `rename` has, not a raw passthrough), `textDocument/inlayHint` and `experimental/externalDocs` (both answer; hover covers enough of the same ground that neither earns a permanent place in every caller's tool list).

Also in here, separately: **`rust_analyzer_symbols` could answer `null`.** The symbols of a file are worked out from that file alone, so that tool deliberately does not wait for the workspace — but between opening a document and rust-analyzer parsing it, the answer is `null`, which reads as a file with nothing in it. Whether the window is hit depends on how busy the machine is: the suite passed run after run here, then failed in `symbols_follow_an_edit` with `invalid type: null, expected a sequence` while several rust-analyzers were competing for the same cores. Exactly the shape of a bug that only ever shows up in CI. It now asks again after waiting, when and only when the first answer was `null`.

Stacked on #42, #43 and #44 — review the last two commits.


zeenix (commented):

Live checks passed for macro expansion, related-tests request handling and runnables; the
symbol retry remains separate from the extension commit. Please also split `9043fa9` into
one commit per independent extension, following CONTRIBUTING.md's atomic-commit rule.

This merges both the #42 and #44 branches, so it inherits the workspace-validation blocker
reported on #42. The symbol null-result fix would be easier to land with the earlier symbol
work, separately from these three new tools.

PR #46, opened 2026-09-12
🐛 mcp: stop one fault being reported, and counted, twice

IconRunner05:
rust-analyzer reports each fault twice: once from its own analysis (`"source": "rust-analyzer"`) and once from `cargo check` proxied through flycheck (`"source": "rustc"`), with different wording each time. The summary counts both, so anything reading the count sees roughly twice the number of faults the compiler sees. On this repo's own `test-project-diagnostics/src/errors.rs` — five errors and two warnings — the tool answered with twenty-three entries.

Same severity, same code, same position is now one entry, with both sources recorded in a `sources` list: a fault rust-analyzer sees and rustc does not is a different thing from one that will fail the build, and collapsing them without saying so would hide that. The wording kept is the compiler's, which names what is wrong with what — "cannot find value `undefined_var` in this scope" against rust-analyzer's "no such value in this scope".

rustc's notes and suggestions are doubled the same way: each arrives inside the diagnostic it belongs to, as `relatedInformation`, and again as a free-standing hint. The free-standing copy carries no context — "value moved here" does not say what was moved — so it goes and the attached one stays. Twenty-three entries become seven, and the summary becomes five errors and two warnings.

The test asserts no two entries share a fault, that the one both sources report carries both and the compiler's wording, and that the summary counts exactly what is in the list.


src/diagnostics/mod.rs
```diff
@@ -14,59 +59,86 @@ pub fn format_diagnostics(file_path: &str, result: &Value) -> Value {
...
         }
+    }
+
+    // rustc's notes and suggestions arrive twice as well: once inside the diagnostic they belong
+    // to, as `relatedInformation`, and once more as free-standing hints. The hint carries no
+    // context of its own -- "value moved here" says nothing about what was moved -- so the copy
+    // attached to its diagnostic is the useful one.
+    let attached: HashSet<&str> = kept
+        .iter()
+        .flat_map(|(diag, _)| related_messages(diag).chain(std::iter::once(message(diag))))
```
zeenix, line 95: Reviewed by Codex on behalf of zeenix.

[P2] Do not include a hint's own message in the duplicate set

`attached` includes `message(diag)` for every kept diagnostic. Consequently every severity-4
diagnostic finds its own message in the set and is skipped, even if it has no related
information and no duplicate. A single rust-analyzer `inactive-code` hint becomes
`diagnostics: []` and `summary.hints: 0` in a focused test.

Only suppress a hint when its message and location are actually represented by another
retained diagnostic's related information; keep standalone hints.

src/diagnostics/mod.rs
```diff
@@ -14,59 +59,86 @@ pub fn format_diagnostics(file_path: &str, result: &Value) -> Value {
...
-        // Count by severity.
-        if let Some(severity) = diag.get("severity").and_then(|s| s.as_u64()) {
-            match severity {
-                1 => errors += 1,
-                2 => warnings += 1,
-                3 => information += 1,
-                4 => hints += 1,
-                _ => {}
+        let key = fault(diag);
+        match faults.iter().position(|seen| *seen == key) {
```
zeenix, line 71: Reviewed by Codex on behalf of zeenix.

[P2] Preserve distinct errors sharing an error code and position

The deduplication key is not a unique fault identity, and this branch collapses diagnostics
even when they come from the same source. For example:

```rust
fn require<T: Copy + Send>(_: T) {}
fn main() { require(std::rc::Rc::new(1)); }
```

rustc 1.98.1 emits two E0277 errors at the same argument span, for missing `Copy` and `Send`.
Passing those two diagnostics through this formatter keeps only the first and reports one
error. Preserve distinct compiler diagnostics, and only merge reports when there is evidence
that they describe the same fault.

zeenix (changes requested):

Two reproduced data-loss defects in the diagnostic formatter: standalone hints disappear,
and separate rustc errors at the same position collapse into one. Both focused regression
tests fail on the current code. Details are inline. This sits on #45 and its dependencies.

PR #47, opened 2026-09-12
🐛 mcp: stop the directory the server started in from reading as dead code

IconRunner05:
#42 made `set_workspace` refuse a directory with no `Cargo.toml`, because rust-analyzer pointed at one loads nothing and answers every question with the `null` that also means "nothing refers to this". It left one door open, which this closes: the workspace a server *starts* in is its working directory, which nothing chooses and nothing checks. A session started outside a Rust project therefore begins in exactly the state `set_workspace` now refuses to move into — and it is the default state for any client that does not pass a workspace argument.

The coverage check added in #42 does not catch it. That one speaks up when the file is outside the loaded root; here nothing is outside it. The root itself is the problem, and every file in it passes the check on the way to an empty answer.

An empty answer from a root with no manifest now says that is what is wrong, and says the root was inherited rather than chosen — which is the part that otherwise reads as a setting somebody picked. Reproduced before the fix: a server started in a directory of dotfiles answered `null` for a `.rs` file sitting in that directory, and would have gone on answering `null` for every question asked of it.

Stacked on #42; review the last commit only.


src/mcp/handlers.rs
```diff
@@ -49,6 +51,90 @@ impl ToolParams {
...
+    server: &RustAnalyzerMCPServer,
+    result: &Value,
+    file_path: &str,
+) -> Result<()> {
+    let empty = result.is_null() || result.as_array().is_some_and(|items| items.is_empty());
+    if !empty {
+        return Ok(());
+    }
+
+    if !server.workspace_root.join("Cargo.toml").is_file() {
```
zeenix, line 108: Reviewed by Codex on behalf of zeenix.

[P2] Do not turn legitimate empty answers into false workspace errors

Reproduced by starting the server in a Cargo project's `src/` directory: hover on a function
returns its correct signature, proving the parent project loaded. Hover on the following
blank line then reports that there is no Rust workspace and that nothing has loaded, solely
because `src/Cargo.toml` does not exist. The same assumption affects `rust-project.json` and
explicitly linked projects.

Base this diagnosis on actual project-loading state/configuration. The absence of a manifest
at the startup directory cannot establish that rust-analyzer failed to load a workspace.

zeenix (changes requested):

The added startup check repeats #42's invalid manifest assumption on another execution path.
A live reproduction is inline. Please fix this check as well as #42's setter validation;
fixing only the setter will leave valid startup configurations broken.

PR #48, opened 2026-09-12
🚸 mcp: say what is at each reference, not just where it is

IconRunner05:
A `Location` is a URI and a range: nine lines of JSON per hit, saying where to look and nothing about what is there. A symbol with thirty references costs a few hundred lines of that, none of which answer the question the list is usually read for — which of these is the one I mean — and each of which has to be resolved by opening a file at a line number.

Each hit is now the file, the position, and the line of source it points at:

```json
{ "count": 3, "locations": [
  { "file": "src/mcp/handlers.rs", "line": 168, "character": 11,
    "text": "    server.ensure_client_started().await?;" } ] }
```

The path is relative to the workspace where the hit is inside it, the coordinates are the ones every tool here takes, so a hit can be asked about without arithmetic, and the whole thing is smaller than the JSON it replaces. Files are read once each, not once per hit.

References only. Definition, implementation and the rest answer with one hit, where the saving would not pay for a second shape for callers to learn.


zeenix (commented):

No additional defect found in the reference-annotation change. Live checks confirm relative
paths, source lines, coordinates and a count matching the returned locations. The fallback
also preserves non-file locations rather than dropping them.

This inherits #46's diagnostic defects and #47's workspace blocker through the current stack.
Reviewing its own commit does not make that cumulative diff ready to merge. #51 adds a
separate declaration-marking defect and should be fixed before combining the two.

PR #49, opened 2026-09-12
✨ mcp: let a call say which workspace it is about

IconRunner05:
The workspace is one setting on the server, and one server is shared by everyone connected to it — an agent and the subagents it fans out, or several sessions of the same tool. Two callers working in two projects each point it at their own, and the loser of that race is answered from the winner's index. Not wrongly enough to notice: a file the loaded workspace does not contain comes back empty, and empty is what dead code looks like. Measured, in the setup this fork is used in: two concurrent callers retargeted the shared server twice, and afterwards a `hover` at a position that had returned a full signature minutes earlier returned `null`.

#42 narrowed that by refusing a workspace with no manifest and by making a file outside the loaded root say so, and #47 closed the same door for the workspace the server starts in. Neither closes it, because the workspaces racing here are all real ones and the files are all inside one of them.

A call may now name the workspace it is about:

```json
{ "name": "rust_analyzer_hover",
  "arguments": { "workspace_path": "/path/to/other-project",
                 "file_path": "src/lib.rs", "line": 10, "character": 7 } }
```

It is answered by that workspace's own rust-analyzer, for that call only, leaving the default where it was for whoever else is using this server. A named workspace is checked the way `set_workspace` checks one, so a typo is a refusal rather than an index of nothing.

One rust-analyzer per workspace asked about, kept for the next question about it. Nothing spawns one implicitly — a call naming no workspace uses the default exactly as before — because each one costs what a rust-analyzer costs. `set_workspace` still means the other workspaces are finished with, so it shuts their rust-analyzers down; naming one per call says the opposite, that it will be named again.


zeenix (commented):

Live testing confirmed that an explicitly named workspace is used and the default workspace
still answers afterwards. No additional blocker found in the routing/cache implementation.

Please fold #50's missing workspace-diagnostics schema property into this feature, so every
applicable tool advertises workspace selection when it first ships. It also inherits the
manifest-validation problem from #42/#47, now factored into `manifest_directory`; propagate
that fix here and into the tool descriptions.

PR #50, opened 2026-09-12
🐛 mcp: let the whole-workspace diagnostic name its workspace too

IconRunner05:
#49 let a call name the workspace it is about, so that a caller sharing this server is not answered from whichever project somebody else last pointed it at. It gave that argument to every tool but one: `rust_analyzer_workspace_diagnostics` declared no properties at all, and still answers for the default. A caller passing `workspace_path` to everything else is therefore answered about the right project everywhere except the one call whose answer is wholly about a project.

It is the only tool asked about no file, which is also why the argument alone was not enough. Every other tool starts rust-analyzer as a side effect of opening the file it was asked about; a workspace named for a single call may be one nothing has opened a file in, so this one starts its own.

Found by a second reader checking the release against its notes rather than taking them — the schema says which tools take the argument, and 18 of 19 was visible there without running anything.


zeenix (commented):

The missing schema property is addressed, and a live check on the combined stack verified
that workspace diagnostics reports the explicitly named workspace. This is best folded into
#49. The inherited diagnostics behaviour still needs #52 and its review fix.

src/mcp/handlers.rs
```diff
@@ -499,7 +878,12 @@ async fn handle_workspace_diagnostics(
     server: &mut RustAnalyzerMCPServer,
     _args: Value,
 ) -> Result<ToolResult> {
-    let Some(client) = &mut server.client else {
+    // Every other tool starts rust-analyzer by opening the file it was asked about. This one is
```
zeenix, line 881: Reviewed by Codex on behalf of zeenix.

[P3] Correct the initialization explanation and remove the redundant call

`dispatch()` already calls `server.ensure_client_started().await?` before matching any tool,
after `handle_tool_call()` has selected the per-call workspace. This handler therefore already
has the correct client even when no file has been opened; workspace-symbol search also has no
file parameter. The schema addition is needed, but the new initialization call and comments
claim a lifecycle gap that the existing dispatcher already covers.

PR #51, opened 2026-09-12
🚸 mcp: say which reference is the declaration

IconRunner05:
A symbol's own declaration is one of its references, so a function with two call sites answers with three hits. That is LSP's behaviour and was equally true of the list of `Location`s #48 replaced — but a list is read hit by hit, where the `count` #48 put at the top of the answer is read as a number, and the number this tool is usually consulted for is how many places use the symbol. Three against two call sites is off by one, in the direction that reads as one more caller than exists. Measured on a function with exactly two call sites: `count` came back 3.

The hit that is the declaration now says so:

```json
{ "count": 2, "locations": [
  { "file": "src/mcp/handlers.rs", "line": 258, "character": 20,
    "text": "    let result = locations::annotate(&result, ...);" },
  { "file": "src/locations.rs", "line": 50, "character": 7,
    "text": "pub fn annotate(locations: &Value, ...) -> Value {",
    "declaration": true } ] }
```

Marked rather than dropped: the declaration is worth having in the list, and a reader subtracting a hit they can see is better served than one handed a smaller number they cannot check. No second count is reported — a reference is not necessarily a call, so a field named for callers would be its own lie — and the description now says what `count` counts.

Which hit it is comes from asking where the symbol is defined, one request against an index that has just answered a harder one. Matched by file and line rather than by column, because a `LocationLink` describes the item where a reference describes the name.


src/locations.rs
```diff
@@ -0,0 +1,115 @@
...
+            "file": shown,
+            "line": line,
+            "character": character,
+            // Same coordinates the tools take, so a hit can be asked about without arithmetic.
+            "text": text,
+        });
+
+        // The declaration is one of the hits and is not a use of the symbol. Saying which one it
+        // is costs a field on a single entry; leaving it unsaid costs every reader an off-by-one.
+        if line.is_some_and(|line| declarations.contains(&(uri_str.to_string(), line))) {
```
zeenix, line 107: Reviewed by Codex on behalf of zeenix.

[P2] Match the identifier range, not just the declaration's line

For `fn f() { f(); }`, both the declaration at column 3 and recursive call at column 9 match
this `(URI, line)` key and receive `declaration: true`. The same happens to a local binding
and its uses on one line, so consumers subtract real uses when interpreting the result.
A focused test with those two locations reproduces the incorrect annotations.

`LocationLink.targetSelectionRange` already identifies the name; compare its full position
(or range) with the reference range, using `Location.range` for ordinary locations.

zeenix (changes requested):

The declaration annotation misclassifies real uses sharing the declaration's line. A focused
regression test fails; details are inline. Also align the PR example with the implementation:
`count` currently includes declaration entries, as the tool description and test require.

PR #52, opened 2026-09-12
🐛 mcp: report the workspace, not whatever happens to be open

IconRunner05:
`rust_analyzer_workspace_diagnostics` asks for `workspace/diagnostic`, which rust-analyzer does not implement — it advertises `workspaceDiagnostics: false` and answers `unknown request` — and then falls back to iterating the documents this server has opened. So the tool answers about whichever files somebody happened to ask about, under a name that says it answers about the workspace.

Where that matters most is a workspace nothing has opened a file in, which since #50 is reachable by naming one:

```
projB, one crate, five errors, nothing opened
  total_files 0, total_errors 0      <- and still zero 45s later
cargo check in projB
  5 errors
```

A confident, correctly-labelled clean bill of health for a project that does not compile. Nothing about the answer looks wrong.

It also under-reports where it appears to work. Two crates, each with two warned-about files, one server rooted in the first:

```
projA (default root, one file opened)   files=1 warnings=1   <- cargo says 2 and 2
```

The file that was opened, not the workspace. A one-file fixture shows the two agreeing, which is presumably how it stood.

A cargo check covers the whole workspace and publishes what it finds for every file in it, opened or not. The report is now the published results of a check this call runs and waits for, asked for with `textDocument: null` — every workspace rust-analyzer holds — which suits the one tool called without a file. Both cases above now match cargo exactly.

Two things fall out of that:

- `complete` says whether the check finished. An empty report with `complete` false is "nothing analysed", a different answer from "nothing wrong", and it now reads as one.
- Each file is counted by the same code as the single-file tool, so a fault that rust-analyzer and rustc both report counts once here as it already did there. The hand-rolled summary this replaces counted both.

Found by a second reader measuring the previous release against two projects rather than one. Stacked on the earlier branches; review the last commit only.


src/lsp/handlers.rs
```diff
@@ -257,17 +390,17 @@ impl RustAnalyzerClient {
...
 
         // The results are published just after the check reports itself done, from a turn of
         // rust-analyzer's loop we cannot see the end of.
         tokio::time::sleep(Duration::from_millis(DOCUMENT_OPEN_DELAY_MILLIS)).await;
 
-        Ok(FreshDiagnostics {
-            items: self.current_diagnostics(uri).await?,
-            complete: loaded && finished,
-        })
+        Ok(loaded && finished)
```
zeenix, line 403: Reviewed by Codex on behalf of zeenix.

[P1] Do not report a failed check as complete analysis

A progress token ending only establishes that the check stopped. rust-analyzer emits the
same end notification when its command fails. Reproduced on a valid Cargo project with
`--config 'check.overrideCommand=["sh","-c","exit 1"]'`: the workspace tool returns
`complete: true`, `files: {}` and zero errors, without any note, although no analysis ran.
A broken check wrapper or similar build-infrastructure failure can hit the same path.

The existing helper's completion heuristic is therefore insufficient for this new workspace
claim. Preserve/inspect check failure state and return an error or `complete: false` when the
check produced no trustworthy report; an ended progress token is not proof of full coverage.

zeenix (changes requested):

The normal case works: a live check of an unopened named workspace reported its errors and
multiple affected files. However, the new completion flag gives a false clean result when
the check fails without analysing anything. Reproduction is inline. The #46 formatter fixes
must also be propagated here because workspace totals now use that formatter.

